PR #8, opened 2016-09-25
WIP

elmobp:
Opening this PR now, once I build the AWS account mapping open source I will add then go for it for now feel free to pull things out of it
- Credentials are hard coded in saml2aws config see README though looking at options to prompt once for them then cache!
- AWS session data is not written to .aws/credentials instead they are set as environment variables and a subshell is launched based on $SHELL
- Command line options to specify an AWS account by name which requires the first step to be done
- Command line options to specify a role to assume
  Only issue which I do not think about ;) is this will only work on Linux/OSX due to the subshell work unless we can get work done to spawn a powershell window?


wolfeidau:
Welcome to open another PR if you complete this work.

PR #16, opened 2016-12-20
ADFS 2.0 (HTTP Auth) support

protomouse:
My understanding is that ADFS 2.0 by default will fall back to HTTP Auth (NTLM/Basic), as opposed to the forms-based authentication used in later versions. That is at least the case on our 2008 R2 installations. [go-ntlmssp](https://github.com/Azure/go-ntlmssp) implements a `RoundTrip` decorator that respects the negotiation scheme used by ADFS 2.0, and implements NTLM authentication. See also: https://aws.amazon.com/blogs/security/how-to-implement-federated-api-and-cli-access-using-saml-2-0-and-ad-fs/.

I've implemented this as a separate "ADFS2" provider. I'm not sure how much sense it makes to merge it with the existing provider.

**TODO:**
- [x] Document usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # saml2aws
 
-CLI tool which enables you to login and retrieve [AWS](https://aws.amazon.com/) temporary credentials using SAML with [ADFS 3.x](https://msdn.microsoft.com/en-us/library/bb897402.aspx) or [PingFederate](https://www.pingidentity.com/en/products/pingfederate.html) Identity Providers.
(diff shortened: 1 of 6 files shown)


wolfeidau:
I think having a seperate provider is a good move, in the longer term we may need to seperate out some common code into a helper.

I think it would be good to change the existing one to add the major version as well, likewise the flags should be `ADFS3` (default) and `ADFS2` I think. Note I am happy to do the refactor once your done.

Thanks for submitting this pull request, I don't have access to an ADFS2 at the moment.

protomouse:
Yeah, my thoughts were similar. I've developed and tested this against our 2.0 installation, and "it works" (by all means do not take my word for it). I've now added docs, and consider this done. I'll leave the renaming/refactoring to you as suggested.

Issue #7, opened 2016-09-25
Hard to identify Account from selection list if using standard role names


mcleane:
saml2AWS dosen't have the account friendly name as part of the role selection list. This can make it hard to select the role you require if standard role names have been used across your aws accounts.
eg

```
[ 0 ]:  arn:aws:iam::123123123123:role/admin
[ 1 ]:  arn:aws:iam::123123123123:role/developer
[ 2 ]:  arn:aws:iam::456456456456:role/admin
[ 3 ]:  arn:aws:iam::456456456456:role/developer
[ 4 ]:  arn:aws:iam::789789789789:role/admin
```

The AWS saml signin page will display the account alias as part of the role selection. 

```
Select a role:
Account: example-cloudops (123123123123)
admin
developer
Account: example-digital-nonprod (456456456456)
admin
developer
Account: example-digital-prod (789789789789)
admin
```

elmobp:
@mcleane have a look at https://github.com/Versent/saml2aws/pull/8 I will see if I can pump out the code Monday for API Gateway + Lambda

wolfeidau:
This is fixed thanks to #22

PR #22, opened 2017-01-30
Account aliases

shawnsi:
This addresses #7 by POSTing to the interactive login form provided by AWS.  Account alias and role name is extracted from the markup.  The CLI prompt has been modified to group by account just like the login form.

I also setup `$ make install` to simplify installation.


wolfeidau:
Nice work that is very helpful!

wolfeidau:
Release is up as 1.5.0 🎉

PR #24, opened 2017-02-26
Add JumpCloud provider

bison:
This adds JumpCloud support based on the existing PingFed provider.

The "hostname" is the IdP URL, e.g.: `sso.jumpcloud.com/saml2/some-application`

Considered omitting the `sso.jumpcloud.com/saml2/` and having the hostname property just be the application name. Could also add support for customizing the prompt on a per-provider basis in that case. Didn't want to do that without asking, but I'm happy to -- also happy to just leave as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,11 @@ Commands:
 ```
 saml2aws will default to using ADFS 3.x as the Identity Provider. To use another provider, use the `--provider` flag:
 
-| IdP          |                    |
-| ------------ | ------------------ |
-| ADFS 2.x     | `--provider=ADFS2` |
-| PingFederate | `--provider=Ping`  |
+| IdP          |                         |
+| ------------ | ----------------------- |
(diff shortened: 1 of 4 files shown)


wolfeidau:
Yes I think the ability to preconfigure profiles is well overdue.

```
saml2aws  configure jumpcloud
url: sso.jumpcloud.com/saml2/some-application
username: mark@wolfe.id.au
provider: ADFS2
Saving alias jumpcloud..
```

Then something like.

```
saml2aws login jumpcloud
Username [mark.wolfe@versent.com.au]:
Password:
```

Then something to migrate the existing configuration.

wolfeidau:
Agree will do a release over the weekend.

bison:
Awesome. Thanks!

wolfeidau:
Release is up as 1.5.0 🎉

PR #25, opened 2017-03-16
setup profile based on user input

valtaz:
I have updated the hardcode user profile 'adsf' to the profile user enter. This will setup a new profile if not exist in aws2saml.config.


wolfeidau:
I kinda did this separately, I owe you lunch.

PR #41, opened 2017-07-07
Add proxy support for KeyCloak

edieship:


wolfeidau:
Looks like I need to do that across all the providers, or provide a common function to create a transport.

wolfeidau:
Thanks for raising it!

PR #42, opened 2017-07-24
fix TLS renegotiation for ADFS

hoegertn:


wolfeidau:
Heya, thanks for the PR.

Just read through https://golang.org/pkg/crypto/tls/#RenegotiationSupport is there a bit of background as to why you recommend enabling it?

Cheers

hoegertn:
I tried using saml2aws with ADFS for a customer and it did not work. It seems that ADFS is doing some TLS renegotiation stuff and go throws errors about this.

After enabling this it worked perfectly.

wolfeidau:
Sounds fair.

Issue #28, opened 2017-04-13
Pass role as CLI flag


sidewinder12s:
Can we add a flag that allows a user to pass in the role they want to assume?

wolfeidau:
Great suggestion, a collegue also has a version which takes the username and pass as flags, i am looking to merge in some of that work. This would be a great addition to that.

sidewinder12s:
We had avoided doing a password flag as it could inadvertently expose a users password in logs.

jghaines:
In addition to this, I'd like to be able to authenticate, assume a parent role and then assume a sub-role in a single command. Is this the same request?:

```
saml2aws
  -a <ADFS account>
  --iam-role-arn arn:aws:iam::123456789000:role/<parent role> 
  --assume-role-arn arn:aws:iam::123456789001:role/<child role> 
  exec -- aws sts get-caller-identity
```

elmobp:
What I've done is save the creds to a file encrypted with gpg, password prompted not via a cli flag I'll publish it next week if anyone interested

PR #43, opened 2017-07-24
add desiredRole cli parameter

hoegertn:
Adds the possibility to specify the desired role as a CLI param.

If only one role is returned it is checked to be the desired role. If a list of roles is returned the desired one is selected.

this implements #28


wolfeidau:
I refactored this whole section and added a ton of new switches including role. I thought I had pushed it as a wip branch but I hadn't :(

I would love some feedback on how I implemented it.

PR #52, opened 2017-08-16
Add additional two factor auth options (SMS and TOTP) to Okta provider

dan-v:
This change adds support for additional two factor auth options for Okta: SMS and TOTP (Google Auth). I added the ability for user to select what two factor auth mechanism they want to use if there are multiple options enabled on the account. Looks like this:

```
Authenticating as user@email.com to Okta https://<otka url>
  1) TOTP MFA authentication
  2) SMS MFA authentication
  3) DUO MFA authentication

Select which MFA option to use: 2
Enter verification code: 036882
Please choose the role you would like to assume:
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The process goes something like this:
 * Identity Provider
   * ADFS (2.x or 3.x)
   * PingFederate + PingId
-  * Okta + Duo
+  * Okta + (Duo, SMS, TOTP)
   * KeyCloak
 * AWS SAML Provider configured
 
--- a/okta.go
+++ b/okta.go
@@ -22,6 +22,20 @@ import (
 	"golang.org/x/net/publicsuffix"
 )
 
+const (
+	IdentifierDuoMfa  = "DUO WEB"
+	IdentifierSmsMfa  = "OKTA SMS"
+	IdentifierTotpMfa = "GOOGLE TOKEN:SOFTWARE:TOTP"
+)
+
+var (
+	supportedMfaOptions = map[string]string{
+		IdentifierDuoMfa:  "DUO MFA authentication",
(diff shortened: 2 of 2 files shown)


wolfeidau:
I will take a look at this soon, thanks for posting a PR!

dan-v:
Thanks @wolfeidau!

wolfeidau:
I have had a good look over this PR and it looks great, my concern at the moment is it adds another question which is asked of the user everytime they log in. Given that this is every hour I am trying to figure out how to make this simpler.

I have messed around with a few ideas but most of them aren't great.

I really need to rewrite the whole flags / config / user input as it is a mess at the moment. I originally considered this a future @wolfeidau problem, sadly that time is now. 

I would love your thoughts on how we can make this smoother. 

If we can't come up with a work around I will just merge your PR as it is great work and others have requested it.

Cheers

dan-v:
Thanks for taking a look at this @wolfeidau. I totally agree with you, and I also was thinking it would be a pain to get an extra prompt every hour. I think ultimately a rewrite of the flags/config/user input would be the best option, but I know that would require a fair amount of work.

The only option I could think of without modifying the flags/config and to avoid the additional prompt would be to assume whatever the first MFA option that is returned is the one that should be used. This definitely falls short especially if there are multiple MFA options enabled and the user doesn't have admin control over the Okta MFA setup.

runningman84:
Does this pull request allow sms auth without using ocata?

dan-v:
@runningman84  - this pull request only adds SMS/TOTP support specifically for Okta.

wolfeidau:
Good news this is merged just working on some tidy up for 2.0 release 😂  🎉

dan-v:
@wolfeidau - awesome! thanks!

PR #76, opened 2017-10-20
feat(ADFS3) Add Symantec VIP MFA support to saml2aws

Anton0:


wolfeidau:
Given you have now tested this I am fine with you merging it.

PR #108, opened 2017-12-11
Feature/command specific help

chamila-c:
Bit of internal refactoring of how flags are handled, but (hopefully) the only user-visible change should be updated `--help-long` output showing which flags are command-specific (vs common to all commands)


cmd/saml2aws/commands/configure.go
```diff
@@ -5,13 +5,14 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/versent/saml2aws"
+	"github.com/versent/saml2aws/cmd/saml2aws/commands/flags"
```
wolfeidau: Would it be better to just move this into a package as in `github.com/versent/saml2aws/pkg/flags`?

chamila-c: Right, will do - I really wasn't sure where it was meant to go, I just put in under commands since that seemed topical to its usage 😄

chamila-c:
@wolfeidau thanks for reminding me, I totally forgot to update the README!
Latest commits should now fix that, as well as move the `flags` package to `pkg` as suggested.
Thanks!

PR #111, opened 2017-12-13
fixed case sensitive imports

### LINKED ISSUE #110: make (glide install) fails
glide.lock line 67 should have `Sirupsen` instead of `sirupsen`, see [Should I use lowercase or uppercase for the package name](https://github.com/sirupsen/logrus/issues/553#issuecomment-306591437) and [Rename back to Sirupsen/logrus](https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276)

same with imports, it needs to match the repo, all imports need to be renamed to `Sirupsen` and `Versent` as well

pete911:
this closes #110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ go:
     - 1.9
 #    - tip
 
-go_import_path: github.com/versent/saml2aws
+go_import_path: github.com/Versent/saml2aws
 
 install:
     - echo noop
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ To use this credential, call the AWS CLI with the --profile option (e.g. aws --p
 
 # Building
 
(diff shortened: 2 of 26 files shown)


wolfeidau:
I need to understand a bit more why you have arrived at this solution, I definitely need to fix the lock file to add logrus.

It seemed the logrus discussion concluded that i should be using the lower case version.

pete911:
orgainization/user on github is called `Versent` not `versent`. When you click on the 'clone or download' button, the link is `Versent` as well ... So it is a bit confusing. I think that the names should be consistent.

wolfeidau:
Yeah I understand the confusion and I will endeavour to fix this in a future change but I am not sure it is critical right now as this is a CLI tool.

I will open another ticket to fix the Versent name in imports.

Issue #114, opened 2017-12-18
ADFS with MFA always fails


pwait:
Looks like it's regressed - no longer asking for a security code, just lets me straight through to a failed login:

```
Response did not contain a valid SAML assertion
Please check your username and password is correct
```

richardkeit:
@pwait - which version, which provider?

pwait:
@richardkeit Provider [0] ADFS and MFA [1] VIP.

The only v2 I am able to use is the linux binaries, at v2.0.0. Windows (and choco, which is a minor behind) binaries don't seem to work at all with ADFS + VIP.

Issue #115, opened 2017-12-18
--mfa flag should validate inputs


chamila-c:
_Affects: v2.0.3_

If the following command is run, the user prompt for "MFA" only allows 2 options, `Auto` and `VIP`:

```
saml2aws configure -a acmeadfs --idp-provider ADFS --username test123 --url https://id.acme.com
```

However if the following command is run, it is possible to enter through the prompts and have the invalid value accepted for MFA, and the incorrect configuration successfully saved (in the `~/.saml2aws` file):

```
saml2aws configure -a acmeadfs --idp-provider ADFS --username test123 --url https://id.acme.com --mfa TEST
```

Similarly, if the `--skip-prompt` option is supplied to the above command, the incorrect value for MFA is still successfully saved.

It would be good if the values supplied to the `--mfa` flag are validated to align with the options provided by the user prompt

PR #119, opened 2017-12-27
fix_vip_mfa_win

marklawler:
- re-enable windows builds
 - remove the default cli option for mfa=auto. If not specifying `--mfa=VIP` on the CLI, it would always default to `Auto` irrespective of what was in `.saml2aws` config
 - put in some validation for provider/mfa combinations

Should resolve issue #114 and #115


wolfeidau:
Thanks for the PR, will give it a shot today.

PR #126, opened 2018-01-24
Add list command to list role ARNs

mspaulding06:
This commit adds a command to list available role ARNs for the authenticated user.  This is helpful to allow the user to list their roles and then they can use the `--role` command line argument to automate exec calls without having to consult the AWS console.

### DIFF
--- a/cmd/saml2aws/commands/list.go
+++ b/cmd/saml2aws/commands/list.go
@@ -1,0 +1,107 @@
+package commands
+
+import (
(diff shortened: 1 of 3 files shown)


wolfeidau:
Works really well, great suggestion.

Would it be better to name it `list-roles`?

Just to be a little more explicit.

mspaulding06:
Makes sense to me, here's an update.

PR #133, opened 2018-01-29
Store password when using the configure command

mspaulding06:
This commit adds a password prompt to the `saml2aws configure` command.  This allows users to store their credentials during configure instead of having to store them the first time they run `login` or `exec`.  This seems to work better for my use case since `exec` calls will be automated and I don't want the user to be prompted at that time.

The user is also asked to confirm their password to ensure it is correct before storing while avoiding authentication.

The password flag has been moved into the common flags struct since it is now used by all commands.

### DIFF
--- a/cmd/saml2aws/commands/configure.go
+++ b/cmd/saml2aws/commands/configure.go
@@ -2,9 +2,12 @@ package commands
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
+	prompt "github.com/segmentio/go-prompt"
 	"github.com/versent/saml2aws"
+	"github.com/versent/saml2aws/helper/credentials"
 	"github.com/versent/saml2aws/pkg/cfg"
 	"github.com/versent/saml2aws/pkg/flags"
 )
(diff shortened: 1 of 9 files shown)


chamila-c:
This change will impact cross-platform consistency until OS password storage functionality is integrated for all platforms.
Until then, could we make the password storage via `configure` an optional feature, rather than always prompting even on platforms that don't support password storage at this time?

mspaulding06:
I think the change I've made meets your requirements.  I've tested on macOS and Linux (Arch) and it appears to be working fine and the password prompt is skipped on Linux.  Let me know if this solution will work for you.

wolfeidau:
@mspaulding06 Can we revert https://github.com/Versent/saml2aws/pull/133/commits/0d2ba0a1c9efb0638eab4b3b84d2b4e0c8c07fea and just add a function to the password provider which returns true or false, depending on whether it supports OS secret storage?

I think this would be easier and keep all that stuff in one place.

mspaulding06:
Made the requested change and put the password prompt into the skip prompt conditional so that it will be skipped as well if the option is enabled.

wolfeidau:
Thanks for making that feature happen, much appreciated.

Issue #135, opened 2018-01-31
Credentials file location inconsistent with awscli on Windows


mberginbh:
On Windows, the default location for the the credentials file is `%USERPROFILE%/.aws/credentials`. 

saml2aws uses `homedir.expand("~/.aws/credentials")` which has slightly different logic when `%HOMEDRIVE%` and `%HOMEDIR%` are set -- see https://github.com/mitchellh/go-homedir/blob/master/homedir.go#L126.

PR #136, opened 2018-01-31
Base default AWS credentials path on USERPROFILE

mberginbh:
Proposed fix for #135

### DIFF
--- a/pkg/awsconfig/awsconfig.go
+++ b/pkg/awsconfig/awsconfig.go
@@ -4,6 +4,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"path"
+	"runtime"
 
 	"github.com/mitchellh/go-homedir"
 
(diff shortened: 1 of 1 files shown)


wolfeidau:
I had a look in the upstream library and it looks like it is already doing this check.

https://github.com/mitchellh/go-homedir/blob/b8bc1bf767474819792c23f32d8286a45736f1c6/homedir.go#L120-L137

There is a PR open to refactor it a bit https://github.com/mitchellh/go-homedir/pull/19

mberginbh:
Upstream has slightly different logic when %HOMEDRIVE% and %HOMEDIR% are set, resulting in inconsistency with awscli in this case. On my system both these and %USERPROFILE% are set, and so awscli and saml2aws have a different idea about where the config file is.

sasmithCalpoly:
Please merge the pull request ASAP. 

The AWS documentation at : https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html
specifies that the default location for .aws on windows is: dir "%UserProfile%\.aws"

Without this being fixed, Windows users with a Home drive mapped will have the saml2aws program putting the .aws\credentials file in the wrong place from where all AWS CLI commands expect the file to be located.

vacovsky:
I fixed the merge conflicts with this PR in a new PR: https://github.com/Versent/saml2aws/pull/270

wolfeidau:
:+1: merged as above.

PR #138, opened 2018-02-11
Added support for RSA tokens with ADFS 2.x

wolfeidau:
This is a pretty unique situation in that the entire authentication flow is different between the default behaviour of the IDP and when it is configured to use RSA MFA tokens.


luckylittle:
Love the branch name :-D

PR #157, opened 2018-04-01
feat(Login) Support sessions up to 12 hours :tada:

wolfeidau:
see https://aws.amazon.com/blogs/security/enable-federated-api-access-to-your-aws-resources-for-up-to-12-hours-using-iam-roles/

* Refactor aws config to use structures
* Added saving of the expiry time for sessions and principal ARN
* Enable content dumping using env DUMP_CONTENT=true while debugging

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,8 +10,8 @@
 [[projects]]
   name = "github.com/PuerkitoBio/goquery"
   packages = ["."]
-  revision = "09540e565986583836b5fc792fe951ec5fd201dd"
-  version = "v1.3.0"
+  revision = "a86ea073017a6beddef78c8659e7224e8ca634b0"
+  version = "v1.4.0"
 
 [[projects]]
   name = "github.com/alecthomas/kingpin"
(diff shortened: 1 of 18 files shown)


wolfeidau:
Once you have updated your roles to permit 8 hour session times, you can use the following options to set this attribute when you login.

To update session time in the default idp account.
```
saml2aws configure --session-duration 28800
```

To do this for a session.

```
saml2aws login --session-duration 28800
```

marianafalzone:
Hi there!
Even when I do this for a session:
`saml2aws login --session-duration=43200`

I still get this message: _Your new access key pair has been stored in the AWS configuration
Note that it will expire at 2019-12-11 10:50:03 -0800 PST_ I logged in at 09:50. This means that even when I specify the duration, it does not work.
Could you please help?

marianafalzone:
Hi @wolfeidau !
I am not sure if I am missing anything but what you suggest is not working for me.
1. I have my roles setup  with a Maximum CLI/API session duration of 12H.
2. I configure all my profiles to run a session of 43200
3. After I run this: `saml2aws login --session-duration 43200` , I get this notification:
_Your new access key pair has been stored in the AWS configuration
Note that it will expire at 2019-12-11 11:43:08 -0800 PST_

Which means that it is still keeping the credentials for 1 hr and I need to extend this time. 

Could you please assist?

PR #164, opened 2018-04-18
Support PingOne and PingID

muramasa64:
It corresponds to the following case.

* Configure SAML setting of AWS to PingOne
* Link PingOne with PingFederate
* Using PingID

### DIFF
--- a/pkg/provider/pingone/pingone.go
+++ b/pkg/provider/pingone/pingone.go
@@ -1,0 +1,500 @@
+package pingone
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/PuerkitoBio/goquery"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/versent/saml2aws/pkg/cfg"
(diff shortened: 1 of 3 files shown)


pkg/provider/pingone/pingone.go
```diff
@@ -0,0 +1,500 @@
...
+	// Redirect from PingFederate to PingOne
+	ac.client.DisableFollowRedirect()
+
+	req, err = http.NewRequest("POST", postUrl, strings.NewReader(formData.Encode()))
+	if err != nil {
+		return "", errors.Wrap(err, "error building authentication request to PingOne")
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	if err != nil {
```
wolfeidau: You have an extra if err check, can you remove this?

muramasa64:
Thank you for your review!
I fixed it.

wolfeidau:
@gliptak Initially I considered this however I am working on some pretty big changes to the pingfed provider to enable authentication without screen scraping. This will only work with pingfed so hence being OK with merging this.

PR #176, opened 2018-05-15
Add global mfa-token flag

Rukenshia:
This flag allows easier automation for running saml2aws with
--skip-prompt. Only Keycloak and ADFS are currently supported.

---
In my exact use case, [saml2aws-auto](https://github.com/Rukenshia/saml2aws-auto), I currently have to rely on passing the MFA token by sending it through `stdin`. This does, however, not work reliably on Windows. I think this addition is quite useful for `--skip-prompt` use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Flags:
       --url=URL                The URL of the SAML IDP server used to login.
       --username=USERNAME      The username used to login.
       --password=PASSWORD      The password used to login.
+      --mfa-token=MFA-TOKEN    The current MFA token (supported in Keycloak, ADFS).
       --role=ROLE              The ARN of the role to assume.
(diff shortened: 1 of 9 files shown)


Rukenshia:
This PR will probably only work if you switch all imports to my fork. I had to do that while developing but obviously switched it back for the PR.

Rukenshia:
Yup, you're right. My bad. just pushed a commit that should fix the test.

Rukenshia:
@wolfeidau do you think this PR is acceptable in this state?

wolfeidau:
@Rukenshia I have it cloned and wanted to try it, been traveling so haven't had a chance to road test!

In my case I need to spin up Keycloak and reconfigure to do it, hence the delay.

Rukenshia:
Sure, no problem. Thank you!

PR #178, opened 2018-06-06
role selection prompt displays sorted list

pete911:
`saml2aws login` always displays role selection in random order, this can be a bit annoying, so added sort - always the same order.

let me know if it is ok, or we should be sorting on account number (or different field)


wolfeidau:
Thanks 👍

PR #179, opened 2018-06-13
fix(pkg/dump) Use httputil.DumpRequestOut, per docs.

emmanuel:
The docs say (https://golang.org/pkg/net/http/httputil/#DumpRequestOut):
> DumpRequestOut is like DumpRequest but for outgoing client requests.

I don't actually see any difference in output when using `DUMP_CONTENT=true saml2aws login -a MY_PROVIDER --verbose`, but the docs seem pretty unambiguous that `DumpRequestOut` is the right API to use here, so I figure it's a safe change.

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -8,7 +8,7 @@ import (
 
 // RequestString helper method to dump the http request
 func RequestString(req *http.Request) string {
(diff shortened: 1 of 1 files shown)


wolfeidau:
Seems reasonable, I didn't know about this, and wanted to see `User-Agent` recently so 👍

Issue #186, opened 2018-06-22
Be smart about form `action` with absolute URL


mjschultz:
Our IdP (Ping), has the absolute URL in the `action` field of the form instead of just a relative URL.

This causes problems on submit because the provider in this package just mashes the [login details URL with the action URL](https://github.com/Versent/saml2aws/blob/954a3b2/pkg/provider/pingfed/pingfed.go#L300).

It looks like the PingOne version might have support for absolute URLs, so it might be easy enough to bring in.

PR #193, opened 2018-07-09
account for absolute URLs in login pages

munkyboy:
fixes https://github.com/Versent/saml2aws/issues/186


wolfeidau:
Thanks a lot!

PR #195, opened 2018-07-12
rework ping federated workflow

munkyboy:
the primary reason for this PR is to add support for ping deployments which utilize the [form.autopost.template.html](https://docs.pingidentity.com/bundle/pf_sm_customizeLocalizeUserFacingPages_pf82/page/concept/idpOrSpUserFacingPages.html) template. The template renders a form which is automatically submitted via javascript.

I tried a few iterations to surgically add support for this but it got messy pretty quick. I ended up using a different model of handling the form workflow madness that is Ping. If the new approach doesn't jive with the style/conventions of this project, please let me know how I might be able to rework this PR. I'm not totally happy with the `context` passing approach for dealing with `loginDetails`. Another approach would of been to attach `loginDetails` to the `Client` struct but that didn't quite seem right either as none of the other providers keep that state.


wolfeidau:
Great work so far, I have been meaning to refactor this but also got bogged down in all the twists and turns of the current approach. 

It will take a me a bit to review this but I will try and do it tomorrow. 

Thanks!

pkg/page/form.go
```diff
@@ -0,0 +1,87 @@
...
+		formFilter = "form[action]"
+	}
+	formSelection := doc.Find(formFilter).First()
+	if formSelection.Size() != 1 {
+		return nil, fmt.Errorf("could not find form")
+	}
+
+	if v, ok := formSelection.Attr("action"); !ok {
+		return nil, fmt.Errorf("could not extract form action")
+	} else {
```
wolfeidau: If you return you won't need an else.

pkg/page/form.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if v, ok := formSelection.Attr("method"); ok {
+		form.Method = strings.ToUpper(v)
+	} else {
+		form.Method = "POST"
+	}
+
+	form.Values = &url.Values{}
+	formSelection.Find("input").Each(func(_ int, s *goquery.Selection) {
+		if name, ok := s.Attr("name"); !ok {
```
wolfeidau: Again this could be trimmed down to remove the else, I would seperate out `name, ok := s.Attr("name")` then do the if statements to "flatten the code".

pkg/page/form_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"bytes"
+	"io/ioutil"
+	"net/url"
+	"testing"
+
+	"github.com/PuerkitoBio/goquery"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewFormFromDocument(t *testing.T) {
```
wolfeidau: Tests are great ❤️

pkg/page/form.go
munkyboy: fixup in 6e3337e

Issue #194, opened 2018-07-12
 Please MFA support of GoogleApps


taru-m:
Hi, @wolfeidau 
It is an error because MFA support is not done.
https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L71

 Please MFA support of GoogleApps.

PR #196, opened 2018-07-13
Japanese correspondence of googleapps

taru-m:
Hi, @wolfeidau

The problem of #194 was caused by not being Japanese-ready.
Please merge.


taru-m:
Thank you!

Issue #121, opened 2018-01-09
Feature: Support for OneLogin as IDP


cjbischoff:
Just looking to see if any plans to support OneLogin for an IDP

pavelnikolov:
Next week I am planning to work on a PR to add support for OneLogin. Any tips would be appreciated.

mumoshu:
@pavelnikolov I'd suggest using the `Generate SAML Assertion` API documented [here](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion). Basically, you can consume id/pass and otp from user input, and then pass it to the API to obtain a SAML assertion, which is then used for [`AssumeRoleWithSAML`](https://github.com/Versent/saml2aws/blob/73cefb0097ea8ee0949aed261af9999294dd6c5f/cmd/saml2aws/commands/login.go#L242).

pavelnikolov:
This comment comes just on time because I literally just resumed my work on this task! Thank you, I’ll review it.

pavelnikolov:
I've got it working. I'll just need to add better test coverage and I'll send a PR. Just wondering what's your preference for provider specific configuration. OneLogin needs 3 additional params `app_id`, `client_id` and `client_secret`. I'm saving `client_*` into the keychain and `app_id` in the config. I have two options - either add interface method that would allow additional provider config (requires some refactoring); or add provider specific config into the common cfg struct (which I don't like).

cjbischoff:
@pavelnikolov Wow thank you - will take a peek

pavelnikolov:
As mentioned in my previous comment, I would like to refactor the code. OneLogin requires four additional params:
* app_id - your application connector id
* subdomain - your organisation unique identifier. Usually used to access OL dashboard https://<subdomain>.onelogin.com
* client_id - Used for generating tokens for authentication with OL APIs.
* client_secret - same as above. I store these two in the keychain. Client ID and client secret are application credentials and not user credentials. 
I’ve added these to the common config and populate them conditionally. Client ID and secret are added as flags during configuration. I would prefer adding app-specific configuration, secrets and flags. However, I wanted to keep this PR minimal.

cjbischoff:
@pavelnikolov Do you have an example of adding/running saml2aws using OneLogin?  I attempted via the prompted method and received the following
```
? Password ***
? Confirm ***
OneLogin provider requires --client_id and --client_secret flags to be set.```

Then I attemped to use the cli option format to add

```
[...]

PR #200, opened 2018-07-18
Add support for OneLogin

pavelnikolov:
This PR adds OneLogin provider support. There are few things that would like to discuss. OneLogin is a bit different than other providers. In order to access their API, each application needs to have `app_id`, `client_id`, `client_secret`. Without these I cannot [generate access_token](https://developers.onelogin.com/api-docs/1/oauth20-tokens/generate-tokens-2), which is in turn used to [generate SAML](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion) assertion and to [verify MFA](https://developers.onelogin.com/api-docs/1/saml-assertions/verify-factor) options.

This PR is very naive and adds the `app_id` into the common configuration and the when prompting for credentials I check if the provider is `OneLogin` and I validate if `client_id` and `client_secret` are provided. Ideally, I would like to refactor the library (i.e. change the signature of the providers) to allow each provider to request/prompt for additional configuration options and keep the common clean. Any feedback and comments would be highly appreciated.

Unit tests will be ready very soon.

### DIFF
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -6,6 +6,9 @@ import (
 
 // CommonFlags flags common to all of the `saml2aws` commands (except `help`)
 type CommonFlags struct {
+	AppID                string
+	ClientID             string
+	ClientSecret         string
 	IdpAccount           string
 	IdpProvider          string
 	MFA                  string
@@ -19,6 +22,7 @@ type CommonFlags struct {
 	SkipPrompt           bool
 	SkipVerify           bool
 	Profile              string
+	Subdomain            string
 }
 
 // RoleSupplied role arn has been passed as a flag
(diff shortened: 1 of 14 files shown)


pkg/flags/flags.go
```diff
@@ -19,6 +22,7 @@ type CommonFlags struct {
 	SkipPrompt           bool
 	SkipVerify           bool
 	Profile              string
+	Subdomain            string
```
pavelnikolov: Ideally, I don't want to put these 4 new fields there. They are only needed for OneLogin. Any ideas where to put them? I wanted to refactor the library and allow each provider to have custom config and secrets (same applies for `ClientID` and `ClientSecret` [here](https://github.com/Versent/saml2aws/pull/200/files#diff-d589c31a93f96430bf47cb898695396aR8).

pavelnikolov:
This PR solves #121

wolfeidau:
So as it turns out I am going to add those attributes for another provider soon to support openid based login.

Although this is a big change and you have touched a bunch of things I do love getting new providers so happy to take it as is.

Cheers for the work you have done!

pavelnikolov:
Subdomain is your organisation unique name. It is usually the subdomain part of the One Login dashboard URL. 

Please, follow this tutorial to generate client credentials https://developers.onelogin.com/api-docs/1/getting-started/working-with-api-credentials

PR #202, opened 2018-07-25
fix(Pingfed) Adjusted selectors to be more lenient.

wolfeidau:
Turns out our implementation of ping doesn't have the same form identifiers as others, to overcome this I have made the code use `pf.pass` to match the login form.


wolfeidau:
CC @munkyboy 

Thanks to your tests I was able to add a fixture which illustrated the breakage and adjusted the code ❤️ 

If you can test with your IDP and report back that would be great, tests pass just want to be sure.

Thanks

munkyboy:
👍 works with my corporate ping deployment

PR #204, opened 2018-07-31
Added package for Shibboleth IDP with Duo support

vacovsky:


vacovsky:
This is only failing because I'm referencing a package that hasn't been merged into the project yet.  Is there a preferred way to add a package and reference it in the same PR?

PR #209, opened 2018-08-20
[OneLogin] Fix duplicate push notifications w/ OLP

pavelnikolov:
The focus of this PR is to fix a bug with OneLogin Protect. The current implementation in the API has a race condition and sometimes it sends multiple push notifications. The issue has been reported to OneLogin and has been fixed. https://onelogin.ideas.aha.io/ideas/IDEAS-I-1588

In order to avoid the issue, we have to add `do_not_notify: true` in subsequent (polling) requests. This is what happens between lines 306 and 311.


pavelnikolov:
Can you, please, release this version?
Thank you!

wolfeidau:
Will do today, just merged in another feature.

PR #210, opened 2018-08-21
feat(secretservice) Add linux secret service support.

wolfeidau:
https://specifications.freedesktop.org/secret-service/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
+sudo: required
 language: go
+os:
+    - osx
 go:
     - "1.10"
 
@@ -10,6 +13,17 @@ install:
 env:
     - "PATH=/home/travis/gopath/bin:$PATH"
 
+before_script:
+    - "export DISPLAY=:99.0"
+    - if [[ "$TRAVIS_OS_NAME" == "linux"  ]]; then sh ci/before_script_linux.sh; fi
+            
 script:
     - make ci
 
+addons:
+    apt:
+        packages:
+        - build-essential
+        - fakeroot
+        - git
+        - libsecret-1-dev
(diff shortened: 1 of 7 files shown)


borgstrom:
Hi @wolfeidau,

This PR is backwards incompatible with previous versions because it forces Linux users to have gnome-keyring and is causing issues for us.

We use saml2aws on our headless AWS instances and installing gnome-keyring there is not something we want to do.  Ubuntu and other distros package gnome-keyring as part of their desktop suites so installing it means that we have to pull in a TON of unwanted dependencies and run a bunch of otherwise unneeded system services (i.e. dbus -- which is the Desktop BUS).

After a quick review of the code I can't see any obvious method for opting out of using the secret service on Linux.  To unblock our users we have forked the project and removed the linux secretservice, but we would like to have a solution in saml2aws and not have to maintain our own fork.

I am requesting that we add an option to saml2aws to opt-out of the secretservice.  We are happy to contribute the code for this if someone could provide guidance on an implementation strategy.  Currently everything happens inside of platform specific `init()` functions, before argv is parsed, so I'm not sure of a straight forward way to opt-out of it.  Otherwise, if you want to tackle it that's definitely appreciated.

From an open source responsibility standpoint I am requesting see some improvements from this project:

* Require all CI tests to pass.  I notice here you've straight-up skipped the tests in the PR in all cases because you couldn't get them to work in Ubuntu.  I submit that this is a direct symptom of the same issue we are facing.  If you can't get it working in CI, your users are unlikely to be able to get it working.
* All PRs should have reviewers tagged and a review should be required to merge.  The fact that a single contributor can operate in a way that causes breaking changes for users is worrying.
* Please maintain a CHANGELOG so users can quickly

Let us know your thoughts!

CC @jfleong 
See also: #219

PR #213, opened 2018-08-24
Fix for changes in duo responses

### LINKED ISSUE #211: DUO MFA response format changes
Yesterday (2018/8/21) we found that Duo changed the format of its MFA response status in a way that broke saml2aws. I have patched it in a fork and would be happy to submit the update here but not sure if the change was global or somehow account specific. Anyone else using Duo care to comment on whether `saml2aws` is now broken for you?

bwhaley:
The response object from duo when polling for MFA status has changed. This PR adjusts the logic accordingly.

Fixes #211


bwhaley:
Looks like #214 is the right fix for this issue so closing this one.

PR #214, opened 2018-08-25
Make additional Okta auth process call if did not poll

themichaellai:
Okta introduced an additional HTTP request in the auth process flow which was addressed in 399beb7. However, in the case that polling does not need to happen, this additional request was not being made.


wolfeidau:
@mcleane @j--wong can one of you take a quick look at this?

themichaellai:
Would someone please be able to look at this? This is currently making us use a slower auth method. :(

PR #226, opened 2018-09-27
add support for role configured in the ini

sstarcher:


emmanuel:
I meant for the 'thank you' to land first, and then the comment on the source code. But I was so eager to check out the code that I read through it first and compulsively commented as I went through. 

Anyhow, the 'thank you' ought to have preceded the comment. Apologies!

sstarcher:
@emmanuel np, updated

sstarcher:
@wolfeidau thank you

PR #231, opened 2018-10-02
dont prompt for creds if they are currently valid

### LINKED ISSUE #227: Only do SAML auth when credentials have expired
It would be nice for a user to be able to run `saml2aws login -a dev` and if the dev credentials have not expired it would not prompt the user.  Also support for `--force` or `--refresh` to forcefully update the credentials if they have not expired.

If this is something you would accept I'll put in a PR for it.

sstarcher:
fixes #227 

I currently use `aws-mfa` and I have a bash function that called switch.  It sets AWS_PROFILE and runs aws-mfa.  If my credentials are expired it will prompt me for my MFA.  If my credentials are still valid it moves along.

Currently with `saml2aws` I have no command I can run that says only reauthenticate me when my credentials have expired.

### DIFF
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -42,6 +42,11 @@ func Login(loginFlags *flags.LoginExecFlags) error {
 		return nil
 	}
 
+	if !sharedCreds.Expired() && !loginFlags.Force {
+		fmt.Println("credentials are not expired skipping")
+		return nil
+	}
+
 	loginDetails, err := resolveLoginDetails(account, loginFlags)
 	if err != nil {
 		fmt.Printf("%+v\n", err)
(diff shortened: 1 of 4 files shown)


wolfeidau:
Great work, when your done ping me and I will drop a release.

Thanks!

sstarcher:
@wolfeidau co-worker ran into one issue and I pushed a fix for that.  Everything for the past day has been working fine for a few of my co-workers.  Looks good on our end.

qcu:
This commit creates a regression, or at the very least does not correctly handle cases where a different account (-a) is specified on the command line.

Ref: #271 

> It looks like the login command should also be validating the role that we have valid credentials for is the same as the one specified in the command line args. This works if the --force arg is included, but I feel like the tool has enough information to determine it needs to re-authenticate.

wolfeidau:
@qcu may be good to add a ticket for this with a couple of examples.

Issue #230, opened 2018-10-01
saml2aws login to work with updated JumpCloud login flow


jefmsmit:
The week of October 8th, 2018 JumpCloud will update the login flow. This previously resulted in issue #224 

Please note that JumpCloud is not explicitly or implicitly supporting saml2aws. The following describes a way that this could work in the future. This too is subject to change.

1. GET Request to get XSRF token
- URL: https://console.jumpcloud.com/userconsole/xsrf
- Response body: {"xsrf": "<Some Long Token Here>"}
- The returned token needs to be used as the value for the `X-Xsrftoken` header on subsequent requests.

2. POST to auth endpoint
- URL: https://console.jumpcloud.com/userconsole/auth
- Headers: 
  - X-Xsrftoken: value is token from previous request
  - Accept: application/json
  - Content-Type: application/json
- Request Body: 
```
{
	"context": "sso",
	"redirectTo": "saml2/aws",
	"email": <user email> string required
	"password": <user password> string required
	"otp": <OTP for MFA> string only required if user has MFA
}
```
`context` and `redirectTo` are required to be the values above.

- If MFA required for user and OTP not provided
  - status code = 401
  - response body : `{"message": "MFA required."}`
- Successful auth
  - status code = 200
  - response body: `{"redirectTo": "https://sso.jumpcloud.com/saml2/aws"}`

3. GET request to `redirectTo` url in auth response to get SAMLResponse

jefmsmit:
These changes can be implemented prior to our release the week of 10/8/2018.

srflaxu40:
@jefmsmit how will users that use jumpcloud retrieve temporary AWS STS tokens without saml2aws for command line use?  Is there an alternative even if with the aws sdk?  From researching it appears the AWS SDK route requires individual MFA protected access for users.

jefmsmit:
@srflaxu40 I am unclear on other options here. I assume someone could implement functionality similar to `saml2aws`. Alternatively, the `saml2aws` community should be able to implement the changes with the steps outlined above.

srflaxu40:
@jefmsmit https://console.jumpcloud.com/userconsole/auth seems to return 404 even when posting on the command-line with the describe headers.  Any advice?

jefmsmit:
Interesting. I can imagine using a GET instead of a POST could cause that, but you said you're POSTing. Can you share a redacted version of your POST body?

srflaxu40:
@jefmsmit - values is the body in my post below:

```
[...]

PR #237, opened 2018-10-16
 Reworked for upcoming JumpCloud login flow change:  #230

rsottilaro:
Followed the guideline that @jefmsmit posted in #230 for what their changed flow will be.

I've tested this against our JumpCloud environment using Ubuntu 18.04, and it appears to work properly for us.


btsteve:
@wolfeidau When will this be in a release?

PR #247, opened 2018-11-12
Add PSU provider

acobaugh:
I wanted to start a discussion around adding a somewhat custom provider to saml2aws. Penn State University has a somewhat customized implementation of Shibboleth+Cosign that prevents us from using the existing shibboleth provider.

I originally went down the road of building our own [tool](https://github.com/acobaugh/psu-saml-aws-login), then discovered saml2aws when researching methods of manipulating the user's AWS profile. This provider is effectively a straight import of the code from my own tool.

Notably, this provider:
* Uses https://github.com/headzoo/surf for web scraping. This is effectively a wrapper around goquery. I could change this to use goquery directly, but it would make the code more complicated
* Has some fmt.Print*'s to allow customizing the output format of some of the prompts, but still uses `prompter` to gather the actual input

I'm willing to change/squash/refactor whatever is necessary if the project owners are willing to accept this PR (I know I have an errant `go fmt` in the commit list, which I can pull out if desired). If not, perhaps we can have a discussion around making the providers more pluggable a-la terraform (whose providers are standalone binaries that the user can drop-in and customize). In the end, I want to be able to take advantage of the framework the saml2aws already provides.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,7 +2,7 @@
 
 
 [[projects]]
-  digest = "1:f8810373a82e746e36ae75d76e979562b19640986c2fc2429e44776c70bd1839"
+  digest = "1:02af1b0c1b08031a940a6c6ad1f55c292c84a7cdd3c435d5c841bcee46819f3d"
   name = "github.com/AlecAivazis/survey"
   packages = ["."]
   pruneopts = "UT"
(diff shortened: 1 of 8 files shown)


acobaugh:
I've refactored some of the code so I could start adding tests, and fixed a bunch of other potential issues. The only linting warning being thrown now is for cyclomatic complexity.

Let me know if you want this PR squashed or any other git hygiene performed. I believe this is ready to be merged.

wolfeidau:
@acobaugh Yes I am back from a week off, could you squash and I will merge.

I have reviewed the PR and it looks pretty well self contained.

acobaugh:
Should be good to go. Squashed, and removed that one fmt commit that touched some of the other providers.

acobaugh:
Thanks, Mark!

Issue #251, opened 2018-11-21
Cannot set a profile other than "saml"  as `--profile`is deprecated and docs are empty


benjackwhite:
Adding the `--profile` flag throws an error saying that it is deprecated and to see "https://github.com/Versent/saml2aws#adding-idp-accounts" but this link goes nowhere.

How can you set the aws_profile to something other than "saml" with --skip-prompt enabled?

acobaugh:
`saml2aws login --profile foo --skip-prompt` works for me. What version were you testing with?

benjackwhite:
Ahhhh okay. I think I understand now. I was giving it to the "configure" command but that is only for the login one. Thanks for the hint!

micahlmartin:
Is there a reason the `--profile` flag isn't used during configure? It can be actually stored in the config so I would think it should be.

acobaugh:
Nope, that seems reasonable.

sagar-srivastava:
I think the issue is that configure command always updates the same profile 'saml' section of the .saml2aws, which inturn will always update the aws cli profile 'saml' which is not desirable in a multi accouint environment. Its better if there is a ways in the saml2aws configure to set the --aws-profile value to something other than 'saml' That way no manual modification of .saml2aws will be needed for multi profiles.

acobaugh:
This should be fixed once #279 is merged.

micahlmartin:
This is now available via the `--profile` option. This issue can be closed.

kwhitlock:
Hi @micahlmartin - I'm not sure the `--profile` option works with configure. I'm running 2.13.0 Still comes up with `unknown long flag`

```
PS Z:\> saml2aws --version
2.13.0
PS Z:\> saml2aws configure --idp-account <omitted> --idp-provider ADFS --mfa Auto --username <omitted> --url <omitted> --aws-urn <omitted> --role <omitted> --profile default --skip-prompt
saml2aws.exe: error: unknown long flag '--profile', try --help
```

Did I misunderstand something? 

Cheers!

PR #252, opened 2018-11-27
Add environment variable support to more flags

acobaugh:
To support a direnv-style project where multiple IDPs or roles may need to be used, add `.Envar` to more flags. Also add the environment variables to help output and perform some minor housekeeping on the help output text.

This also updates the link in the help output, identified in #251 as being broken.


wolfeidau:
Thanks!

PR #259, opened 2018-12-20
Add a configurable option to set the duo mfa option

srizzling:
This change adds a configurable option so that users can set a flag to configure their prefered duo option (Push or Passcode). This is useful for users, who setup aliases, and want a swift and easy approach to login into the accounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,23 +16,7 @@ The process goes something like this:
 
 ## Table of Contents
 
-<!-- TOC depthFrom:2 -->
-
-- [Table of Contents](#table-of-contents)
-- [Requirements](#requirements)
-- [Caveats](#caveats)
-- [Install](#install)
-    - [OSX](#osx)
-    - [Windows](#windows)
-- [Dependency Setup](#dependency-setup)
-- [Usage](#usage)
-    - [`saml2aws script`](#saml2aws-script)
-    - [Configuring IDP Accounts](#configuring-idp-accounts)
-- [Example](#example)
-- [Building](#building)
-- [Environment vars](#environment-vars)
-
-<!-- /TOC -->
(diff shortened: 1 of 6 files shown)


pkg/creds/creds.go
```diff
@@ -9,6 +9,7 @@ type LoginDetails struct {
 	Username     string
 	Password     string
 	MFAToken     string
+	DuoMFAOption string
 	URL          string
```
pwmcintyre: I like to think this could be just MFAOption, leaving it open to be implemented on other providers?

pkg/provider/okta/okta.go
```diff
@@ -371,7 +371,15 @@ func verifyMfa(oc *Client, oktaOrgHost string, resp string) (string, error) {
 			"Passcode",
 		}
 
-		duoMfaOption := prompter.Choose("Select a DUO MFA Option", duoMfaOptions)
+		duoMfaOption := 0
+
+		if loginDetails.DuoMFAOption == "Duo Push" {
```
pwmcintyre: I'm not sure this is safe, I don't think there's any certainty of the order - you may need to iterate over each and compare

srizzling: I assumed it was OK, since the code previously assumed an order (and since the array is declared just above)

I could iterate to be make it safer however.

README.md
```diff
@@ -16,23 +16,7 @@ The process goes something like this:
...
-- [Dependency Setup](#dependency-setup)
-- [Usage](#usage)
-    - [`saml2aws script`](#saml2aws-script)
-    - [Configuring IDP Accounts](#configuring-idp-accounts)
-- [Example](#example)
-- [Building](#building)
-- [Environment vars](#environment-vars)
-
-<!-- /TOC -->
+<!-- TOC depthFrom:2 -->autoauto- [Table of Contents](#table-of-contents)auto- [Requirements](#requirements)auto- [Caveats](#caveats)auto- [Install](#install)auto    - [OSX](#osx)auto    - [Windows](#windows)auto- [Dependency Setup](#dependency-setup)auto- [Usage](#usage)auto    - [`saml2aws script`](#saml2aws-script)auto    - [Configuring IDP Accounts](#configuring-idp-accounts)auto- [Example](#example)auto- [Building](#building)auto- [Environment vars](#environment-vars)autoauto<!-- /TOC -->
```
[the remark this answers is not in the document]
srizzling: I fixed it.. I think I'll remove those TOC comments since different plugins would interact with it differently.

PR #277, opened 2019-01-31
Adding documentation for JumpCloud

nshenry03:
I thought that this might be helpful for some

### DIFF
--- a/doc/provider/jumpcloud/README.md
+++ b/doc/provider/jumpcloud/README.md
@@ -1,0 +1,215 @@
+# saml2aws Documentation for JumpCloud
+
+Instructions for setting up single sign on (SSO) with Amazon AWS using
+[JumpCloud][1] and [saml2aws][2].
+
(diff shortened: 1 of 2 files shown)


doc/provider/jumpcloud/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Based on the [instructions from JumpCloud][4], we'll generate a public
+certificate and private key pair for administrative access to our production
+AWS account.
+
+> I believe it's possible to use the same certificate and key for all SSO
+> applications; however, I think it is best practice to generate a pair for
+> each application.
+
+Create `production.cnf`:
```
nshenry03: I believe it's possible to use the same certificate and key for all SSO
applications; however, I think it is best practice to generate a pair for
each application.

wolfeidau:
I :heart: documentation so thanks for adding that!

Issue #267, opened 2019-01-08
Powershell environment variables are missing quotes and has an equals typo


Scoth42:
When doing a saml2aws script with the --shell=powershell, the resultant variables are missing their quotes (except the last one) and there's a double quote in AWS_SECRET_ACCESS_KEY. I don't have a Windows compliation environment handy for testing else I'd submit a pull request myself, but it should be an easy fix.

https://github.com/Versent/saml2aws/blob/100b384b9efe95955658f3d148f5eea51f4fb1cf/cmd/saml2aws/commands/script.go

PR #281, opened 2019-02-12
add quotes remove extra equals from powershell env script

TylerReid:
These changes are to address this issue: https://github.com/Versent/saml2aws/issues/267
* removed extra quote in setting secret key
* added single quotes around the values to treat as literal strings (shouldn't execute a profile name of `$(Remove-Item "C:\Windows" -Force)`)


TylerReid:
As an aside, I think the bash template should also have single quotes unless I am missing some reason you would want a profile name that is expanded. https://github.com/Versent/saml2aws/blob/master/cmd/saml2aws/commands/script.go#L14

Issue #248, opened 2018-11-15
Feature Request: Support RSA App/Hard Token


akr257:
Hi,
The current version works fine when a user has a SMS (RSA) token. The current behaviour is:

1.  User inputs their 4-8 digit RSA PIN
2.  User receives an SMS token on their mobile phone - user enters this code 
3.  User is logged in

This however fails if the user has a App Token (RSA Mobile App) or a hard token as the user log in process is different:

1. User inputs their 4-8 digit PIN + token on the RSA Mobile App (or Hard Token)
2. User is logged in

Can the app have an option to specify which version of the RSA token they have (during saml2aws configure) and then present the appropriate login option. 

Thanks.

PR #286, opened 2019-02-28
Add check for SAML response after first RSA code submission

ethackal:
Adding a check for a SAML response after the first RSA code is submitted enables hard and virtual tokens which do not normally require nextCode. This should fix https://github.com/Versent/saml2aws/issues/248. The current tests do not include a page for nextCode so I cannot properly test that it does not break prior behavior.


suaswe:
Hey ethackal/Versent! Is there any chance of this being merged in? I'm hoping it will fix my inability to authenticate using MFA.

PR #295, opened 2019-03-29
- duo-mfa-option doesn't work, or is no longer an option

jasonghelfman:


jasonghelfman:
closed and created with different account

Issue #299, opened 2019-04-03
Missing flag / ability to set location of .saml2aws config file.


keirwhitlock:
The NewConfigManager func takes in a configFile string, and if this is  "" sets it to the one in the constants as per: 

https://github.com/Versent/saml2aws/blob/c62630b8472888e5090a9ba27c2624882518c830/pkg/cfg/cfg.go#L130

In the calling of the func we are passing in the const in the cfg anyway, so seems hardcoded to "~/.saml2aws", as shown here:

https://github.com/Versent/saml2aws/blob/2771aa606604608981e59a2a9d912f89d7f43611/cmd/saml2aws/commands/configure.go#L25

I feel like the func call should have a user inputted location of a configFile like that with the 'IdpAccount', provided by a cmdline flag, and if none was provided default to the constant.

Was this the desired intention, and as yet not implemented? Or have I missed something here? (I'm quite new to Go so that's entirely possible :) )

Cheers!

PR #303, opened 2019-04-08
Passing in configFlag to NewConfigManager() instead of the const

kwhitlock:
This is my attempt at solving #299


kwhitlock:
I think the Travis check has stalled? It's green on the details page, but still show as waiting here. I think it needs restarting...?

wolfeidau:
Thanks :+1:

PR #309, opened 2019-04-23
Azure MFA Support

shawnsi:
This PR implements support for Azure MFA when using ADFS.  We are testing this internally for my enterprise use case and thought the community may benefit from inclusion.

Please let me know if there are any questions or changes I can make to encourage merging this PR.


shawnsi:
After discussion with Microsoft they are discouraging any approach that scrapes the Azure MFA forms.  I'll be continuing to work with them to describe the use case and hopefully find an enterprise solution that fits.

PR #314, opened 2019-05-29
New provider: ShibbolethECP

acobaugh:
This new provider supports the Shibboleth ECP profile, designed for non-browser authentication (with optional Duo) against Shibboleth.

I'm planning on adding unit tests, they just might not make it in before this PR is reviewed.

(the backstory is we at PSU are switching to the Duo web API for our single-signon which was going to be nearly impossible to screen-scrape as we were previously doing with the PSU provider. Shib ECP bypasses all of that, albeit with SOAP-XML)


acobaugh:
I think this is good to go now. Can we get this merged? @wolfeidau

PR #317, opened 2019-06-03
Add Support for Shibboleth idp-provider in cli

maximede:
to allow `saml2aws configure -a saml --idp-provider=Shibboleth`


wolfeidau:
👍

Issue #311, opened 2019-05-03
Google Apps unsupported second factor


wdalmut:
Hi,

i'am trying out this project,. On our google account we use the u2f as second factor auth and `saml2aws` is unable to generate temporary credentials and fails with this error:

```
error authenticating to IdP: error loading challenge page: unsupported second factor: /signin/challenge/sk/6
```

Anyone have the same problem?

Bests
Walter

wdalmut:
Without the u2f technology as mfa everything works correctly

jstangroome:
I have a very similar error:
```
error authenticating to IdP: error loading challenge page: unsupported second factor: /signin/challenge/sk/5 
```

My Google Account 2FA is configured to use my Yubikey 5 NFC by default.
![image](https://user-images.githubusercontent.com/148754/57979665-674e1700-7a64-11e9-98b3-7df644c92ef8.png)

Issue #305, opened 2019-04-15
Choose different MFA on Google Auth if the main fails


paulopontesm:
Got the error `Device can't be reached right now` because of some problem in my device. 

![image](https://user-images.githubusercontent.com/3610528/56140440-d54d7d00-5f92-11e9-82ed-1e150da610d5.png)

It would be helpful if it showed an option to choose a different 2FA method.

Issue #238, opened 2018-10-17
Support Yubikey challenge/sk for Google login


dncarley:
Need to support `challenge/sk` for Google Apps Login: https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L154

so that users can use a [Yubikey](https://www.yubico.com/) for multifactor authentication.

ahrenstein:
As as stop gap I'd be happy if we could set it to force ToTP as if it were a browser that doesn't take U2F.

mothershipper:
Got this working at #398, it could probably use a second pass since I did most of it late last night - I won't have a chance to take a second look until later this week.

george-angel:
Hello, is there any possibility to push this work forward?

I noticed that https://github.com/marshallbrekka/go-u2fhost/pull/10 has been merged, which I think is relevant.

Thank you.

rayterrill:
Any word on whether this works yet, or anything that we can help with in getting this implemented?

PR #321, opened 2019-07-02
Try alternate Google 2FA method if primary isn't supported

idubinskiy:
Implements #305

Workaround for #238, #311


idubinskiy:
@wolfeidau Thanks for the merge! Any idea when a new release will be tagged/available on the Homebrew cask?

Issue #146, opened 2018-02-26
Compatibility with Windows Azure Active Directory (WAAD)


csghuser:
Has anyone switched to using Azure AD to provide single sign on and managed to get this working?

For reference the guide is here:

https://docs.microsoft.com/en-us/azure/active-directory/active-directory-saas-amazon-web-service-tutorial

Is it possible with the current code or would an additional provider have to be written?

davidobrien1985:
@wolfeidau would need to correct me, but I believe there would need to be some extra code that needed to be written. Especially because the AAD prompt needs to be caught by a screenscraper essentially.

ashemedai:
Hi, I got an interest in getting this working. Is there anything aside from the `pkg/providers` that I need to work on to get it working?

I assume calling it `waad` is good enough or do we prefer `azuread`?

davidobrien1985:
Either `azuread` or `aad`, my preference would be the former.

ashemedai:
Just a small heads up: working on this a small bit at a time. Working on the scraper at the moment.

ashemedai:
@tomcroll Thanks for the offer. I'll try to submit a first PR next week. Was sidetracked with some SessionDuration custom claim and other work stuff.

brettneese:
@ashemedai did you get anywhere with this? :) I was about to hack this together until I saw this issue so happy to help if I can.

brettneese:
totally understandable! :) thanks for posting that.

st33v:
has anyone been able to get this working? I'm totally stuck.

steven-cuthill-otm:
just seeing if there has been any progress on this ?

jaxxstorm:
If anyone sees this, I am happy to sponsor the development of this feature - please contact me.

rdkls:
There's also this https://www.npmjs.com/package/aws-azure-login - however I (currently) can't get it to run non-interactively. Would love to just keep using saml2aws though.

paihu:
Hello I wrote the support of AaureAD. It seems to work for my AzureAD tenant but I could not test other environments.

https://github.com/paihu/saml2aws/commit/572b549948eb310aa3008e7c2c280cc60c5a9401

pysysops:
I seem to get a login webpage returned 🤔along with:
```
DEBU[0010] HTTP Req                                      URL="https://login.microsoftonline.com/common/login" http=client method=POST
DEBU[0011] HTTP Res                                      Status="200 OK" http=client
[...]

PR #325, opened 2019-07-25
 New Provider: AzureAD

paihu:
see #146 

It seems to work for my AzureAD tenant but I could not test other environments.


chrisbrownie:
@paihu when configuring a profile, what do you use for URL? If you can give me some details, I'll get the docs updated so that others can actually use this.

PR #328, opened 2019-08-01
Do not print spinner if being redirected

worr:
In cases where saml2aws is being redirected to a file, or where customers are
using saml2aws as a library, it can cause problems to output the spinner when
stdout is not a tty.


worr:
An alternate way to resolve this would be to move the spinner code to the `cmd` portion of the tool, and just start spinners when we want to display them to users. I'd be happy to implement that if y'all prefer that.

wolfeidau:
I think this is a good fix and nice and specific in it's current location.

Issue #335, opened 2019-08-19
Ability to run `saml2aws exec` against a specific AWS Profile


85matthew:
As a user I need to be able to utilize my local ~/.aws/config file in order to access various AWS accounts.

saml2aws currently does not appear to support this functionality. It will only allow you to pass a `profile` tag which needs to match the name of the profile name specified when running `saml2aws login`. (Without specifying, "SAML" is the default profile where the temporary credentials are stored)

After obtaining an initial temporary set of credentials to access a "hub" AWS account, (via `saml2aws login) I need the ability to assume-role into other "spoke" AWS accounts that are connected to my hub account.

Proposing a new flag in order to avoid breaking any existing functionality. Can utilize a flag called '--exec-profile'

85matthew:
I have a proposed PR to achieve this functionality. I would be glad to open a PR and get feedback.

I did see a couple other open issues that may be related but I was not confident based on the descriptions.

wolfeidau:
Fixed by #336

PR #336, opened 2019-08-19
#335 Proposal to allow `saml2aws exec` against specific AWS profile

85matthew:
Proposal for discussion regarding issue #335 

See README for how to utilize the proposed flag.


wolfeidau:
This is a great suggestion, personally I use saml2aws aliases to define more than one entry for the same account to use different aws profiles. For example `myaccount-admin-syd` alias in saml2aws which logs in and uses `myaccount-admin-syd` profile in `~/.aws/config`.

wolfeidau:
I have added https://github.com/Versent/saml2aws#advanced-configuration to illustrate multi account / profile / role setups. 

Is this the kind of extra documentation you were after?

85matthew:
This documentation is helpful for people with roles in multiple accounts that want to SAML authenticate against every account like this.

The PR that I did works a bit different. We only SAML authenticate to a single AWS account. Once authenticated and assigned a role from the single AWS account, we want to be able to assume the role in a different AWS account in order to execute commands via aws-cli {Without SAML against the 2nd (or other remote) accounts}

Our environment looks like this:

                                --------------------------- - ----          
SAML2AWS --->  |  'Authorization' AWS Acct            |
                               |   (Provides only single point        |
                               |   (of mgmt/monitoring/entry|      |
                               |    no other services in this acct  |
                                ---------------------------------
                               |                           |
                               |                           |    Assume Role (No SAML) that can act upon other AWS acct.
                               |                           |      'exec-profile' would let me use a 'prod' role via assume-role 
                               |                           |      from the SAML 'Authorization' AWS account
          ----------------------    --------------------
          |  Remote Acct (prod)|   | Other Acct (stag) |
          -----------------------  --------------------

As you can see, I only have a single point for SAML authentication (the 'SAML' profile in my ~/.aws/config file) The PR that I did allows a command to be executed with a role that is assumed from the SAML entry point.

so something like:

saml2aws login   (\# To authenticate against my 'Authorization' AWS account)
saml2aws exec --exec-profile prod aws s3 ls     (Would list s3 buckets in 'prod' account)
saml2aws exec --exec-profile staging aws s3 ls     (Would list s3 buckets in 'staging' account)

with this in my ~/.aws/config file

[profile SAML]  (\# SamlRole in 'authentication' account)
region=us-east-2
output=json
 role_session_name=user.name

[profile prod]
 source_profile=saml
 role_arn=arn:aws:iam::00000002:role/ProdRole (\#Prod account)

[profile staging]
 source_profile=saml
 role_arn=arn:aws:iam::00000003:role/StagRole (\#Staging account)



Hopefully that helps explain. Let me know if it doesn't make sense. Thanks @wolfeidau !

85matthew:
wow. That mangled by ascii drawing.

wolfeidau:
If you add your use case to the README under the section I added I am happy to merge this.

85matthew:
Thanks for getting back Mark. I’m traveling this week but will touch base
with you next week when I’m back. Thanks and talk soon.

PR #338, opened 2019-08-22
add Void Linux install instructions

ScOut3R:
The package was just added to Void Linux. Someone might find it handy (I do ;)).


README.md
```diff
@@ -71,6 +71,16 @@ choco install saml2aws
 saml2aws --version
 ```
 
+### Linux
+
+#### Void Linux
+
+If you are on Void Linux you can use xbps to install the saml2aws package!
```
wolfeidau: Can you make Void Linux a link to the project?

ScOut3R: Sure, it's done.

PR #343, opened 2019-09-18
Fix to Support ADFS5 Form Format

### LINKED ISSUE #283: Connecting to ADFS
As it seems, there are some configuration possibilities to login to ADFS: ADFS either ADFS2 and there are also sub-options which I do not know.

A couple of configuration ways I consider default do not work (login fails). That is, ADFS and ADFS2 with "saml/Auto".

```
Authenticating as xxxxx...
Response did not contain a valid SAML assertion
Please check your username and password is correct

```
All I do know is that we have ADFS2016 and login works via UI. How to find out/what to ask the system administrator to know what options are correct?

nathandines:
Fixes #283

ADFS 5.0 on Windows Server 2019 introduces a hidden field with the name
of "UserName" and the id of "userNameInputOptionsHolder". saml2aws will
include this in the fields which it populates when submitting a request
to ADFS.

ADFS encounters an error when both "UserName" fields are populated, so
this change ensures that only "user" and "password" fields which are
visible to the user will be populated by saml2aws and submitted to ADFS.

### DIFF
--- a/pkg/provider/adfs/adfs.go
+++ b/pkg/provider/adfs/adfs.go
@@ -185,13 +185,23 @@ func updateFormData(authForm url.Values, s *goquery.Selection, user *creds.Login
 	if !ok {
 		return
 	}
+
+	typeValue, typeFound := s.Attr("type")
+	hiddenAttr := typeFound && typeValue == "hidden"
+
 	lname := strings.ToLower(name)
 	if strings.Contains(lname, "user") {
-		authForm.Add(name, user.Username)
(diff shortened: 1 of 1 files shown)


nathandines:
Actually, going to pause on this. Might need a little more work

nathandines:
Hey @wolfeidau, any chance we could please get this into a release soon?

I'm hoping to avoid distributing a custom build at one of my clients and keep them using the proper installation channels.

PR #361, opened 2019-10-10
Add disable-keychain option and implementation

skystar-p:
Someone who uses [pass](https://www.passwordstore.org/) or other password managers to manage their credentials, storing credential into keychain can be an inconvenient option. This patch provides `--disable-keychain` option which disables behavior that storing credential into their local keychain.


serialx:
Also, when one uses saml2aws in a ssh remove environment, it is literally unusable. This option would resolve my issue.

wolfeidau:
@skystar-p I would prefer `disable` over `no` as in `--disable-keychain` and the addition of an environment variable option for this of `DISABLE_KEYCHAIN` if that isn't too much trouble.

Otherwise 👍

wolfeidau:
Cheers for building this feature 🎉

PR #362, opened 2019-10-11
Update exec to perform role assumption from profile.

elementalvoid:
The method created by @85matthew worked for a large majority of tools
but we found that some don't properly utilize the AWS_PROFILE env var.
In particular the Sops tool has issues.

This update makes it such that saml2aws behaves more like some other
tools in this space ([aws-vault](https://github.com/99designs/aws-vault) and [aws-okta](https://github.com/segmentio/aws-okta)). When the `--exec-profile`
flag is specified saml2aws now performs the role assumption via the STS
client from the AWS SDK. This means it will now do multiple levels of
role assumption _before_ executing the given command. Because saml2aws
is performing the assumption it no longer exports the AWS_PROFILE env
vars and the credentials that it export are fully fledged credentials
for the target profile.

BTW: @85matthew and I work together so this update doesn't break anything he was trying to achieve with #335.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/alecthomas/units v0.0.0-20190910110746-680d30ca3117 // indirect
 	github.com/andybalholm/cascadia v1.0.0 // indirect
 	github.com/aulanov/go.dbus v0.0.0-20150729231527-25c3068a42a0 // indirect
-	github.com/aws/aws-sdk-go v1.14.8
+	github.com/aws/aws-sdk-go v1.23.15
 	github.com/beevik/etree v1.0.1
 	github.com/briandowns/spinner v0.0.0-20170614154858-48dbb65d7bd5
 	github.com/danieljoos/wincred v1.0.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/alecthomas/units v0.0.0-20190910110746-680d30ca3117 // indirect
 	github.com/andybalholm/cascadia v1.0.0 // indirect
 	github.com/aulanov/go.dbus v0.0.0-20150729231527-25c3068a42a0 // indirect
-	github.com/aws/aws-sdk-go v1.14.8
+	github.com/aws/aws-sdk-go v1.23.15
```
elementalvoid: I had to upgrade the SDK in order to get this to work. However, I did not upgrade to the latest version, only to a version I knew to work. If desired I'll bump to `v1.25.10` which [as of yesterday is the latest release](https://github.com/aws/aws-sdk-go/releases).

kr3cj: For reference, `v1.14.8` was released in June of 2018 and `v1.23.15` was released last month.

kr3cj:
This looks good to me. Just wondering if we should clarify any documentation or READMEs on the enhancement.

wolfeidau:
Bumping the SDK is well overdue, thanks!

elementalvoid:
@wolfeidau thanks for the merge, happy to help since you all have given us a great tool!

elementalvoid:
Hey @wolfeidau -- Thanks again for merging this change; it's going to be crucial in our multi-account AWS journey.

Might you have a guess as to when a new release will be created with this change included? For the short term I can provide our teams a local binary but allowing them to install/upgrade via Homebrew would be nice to have.

wolfeidau:
I have cut a new release 🎉

PR #365, opened 2019-10-21
External provider

adammw:
Add new "external" provider that shells out to a separate script to retrieve the SAML response.
Uses the URL field as the command to pass to /bin/sh to execute.

Tested locally with an external script that hijacks the existing browser session to retrieve the SAML response, however could be used with anything really.

### DIFF
--- a/pkg/provider/shell/shell.go
+++ b/pkg/provider/shell/shell.go
@@ -1,0 +1,30 @@
+package shell
+
+import (
+	"os/exec"
+
+	"github.com/sirupsen/logrus"
+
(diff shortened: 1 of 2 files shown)


wolfeidau:
@adammw I ❤️ this PR, however I would prefer to call it a "shell provider". I was planning to build a plugin system at some point so external was on my list of names for RPC to another program kind of like hashicorp does in packer ect.

Cheers

adammw:
Renamed to Shell Provider @wolfeidau

PR #366, opened 2019-10-21
Add support for parsing AWS Config file to automatically select role

adammw:
e.g. if `~/.aws/config` specfies:
```
[profile production]
role_arn = arn:aws:iam::123456789:role/admin
```

then running `saml2aws login --profile production` will not prompt for a role and will automatically log in to the chosen role, similar to if it were set for the IdP config.

### DIFF
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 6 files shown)


adammw:
cc @wolfeidau

wolfeidau:
I really like this feature, I wonder if it will confuse users, almost needs a prompt and a flag to just use last role.

Just concerned it may annoy people who switch roles, I thank that 99% of users will find this super helpful through. 

Thoughts?

elementalvoid:
First let me say that I'd love to have this feature. I have shell aliases in place for my different roles to use the appropriate `login --role` value. I could swap to specifying profile names as opposed to role ARNs which would be nice.

However, there is already a `--profile` flag on the `login` command. It sets which AWS profile to store the SAML credentials into. I, as one data point, do utilize this flag sometimes. It's probably an abnormal use case but I store SAML creds into several AWS profiles then I use the `source_profile` on additional AWS profiles to pull from those as needed. This allows me to have several active SAML sessions with different permission sets.

When my teammate and I added the `--exec-profile` flag to the `exec` command we found that there too was already a `--profile` flag which sets which profile to pull the SAML credentials out of.

We added the `--exec-profile` to avoid a collision on this pre-existing flag. I suspect the same will need to be done here? Perhaps something like `--role-profile`?

adammw:
Yeah I had to back out this change locally, as it kept trying to assume the role it already was. Will need to go back to the drawing board but haven't had the time.

mapkon (changes requested):
Hello @adammw 

Is this PR still relevant? If yes, can we reconcile it to the latest master? If not, I am considering closing it.

Thanks

mapkon:
Closing this - there has been no traction on it.

Issue #118, opened 2017-12-26
Please help on Okta MFA


kevbook:
Sorry, I never ask these questions on repo issues. [`v2.0.3` on mac 10.13.2]

Config for multi-account done exactly as here http://saml-doc.okta.com/SAML_Docs/How-to-Configure-SAML-2.0-for-Amazon-Web-Service.html

* When I setup Okta with NO MFA - things work just fine. 
* But with MFA setup on AWS (google authenticator) - I do `saml2aws login --verbose` or `saml2aws login --verbose --mfa="totp"`

```
Response did not contain a valid SAML assertion
Please check your username and password is correct
```

wolfeidau:
Have you tried disabling MFA and logging in?

Just to ensure your basic configuration is working.

kevbook:
Oh yea basic works just fine with multiple AWS accounts, using `saml2aws` & okta web.
But `saml2aws` gives me the above error when MFA is enabled

chamila-c:
Any chance this is related to https://github.com/Versent/saml2aws/issues/84 ?

kevbook:
Seems like when MFA on Okta is setup on okta org level, this works. But not if MFA is setup at the AWS app level.

kevbook:
See this as well. https://github.com/Nextdoor/nd_okta_auth/issues/2

etendards:
Is this still an issue?  How do we enable MFA with CLI?

wolfeidau:
OK I need to fire up my okta developer account and take a look at the output of --debug. 

I didn't even know you could do per app configuration of MFA.

wolfeidau:
@pwmcintyre How are you configuring this at the moment?

emmanuel:
I'm experiencing this same issue with 2.5.0 on macOS; my organization has Okta configured for MFA on a per-application basis (i.e., MFA is enabled for AWS, but not all apps in the tenant). (Based on my understanding of the underlying issue, neither the version of `saml2aws` nor my OS are relevant, however.) 

I have cycles to prepare a patch for this functionality, but I think I'll need a little guidance to do so effectively. Can someone help me sort out how to diagnose what Okta does differently in the per-application case vs. in the tenant-wide case?

emmanuel:
... looks like I have to rescind my offer here; I was mistaken about having cycles for this. My organization is switching to org-wide MFA tomorrow, so there's no justification for investing dev effort in support of app-level MFA. 

Apologies for the offer-then-retraction.

[...]

PR #369, opened 2019-10-28
Fixes Okta MFA so that it works with TOTP

alsmola:
I wasn't able to get `saml2aws` to work with Okta using MFA for our organization. We have the "Okta Verify" MFA (in reality just a TOTP scheme) configured for our AWS account and I was getting the "response does not contain a valid SAML assertion" error from https://github.com/Versent/saml2aws/issues/118.

According to [Okta support](https://support.okta.com/help/s/article/Integrating-the-Amazon-Web-Services-Command-Line-Interface-Using-Okta) the "preferred solution" is gimme-aws-creds. I looked at [their code](https://github.com/Nike-Inc/gimme-aws-creds/blob/master/gimme_aws_creds/okta.py#L543-L548) and see they grabbed the state token from the page and resubmit to the `authn` endpoint.

Using that approach, I was able to get `saml2aws` to ask for my MFA code and return a valid AWS session.

Happy to update or add any additional comments/documentation. Thanks!

### DIFF
--- a/pkg/provider/okta/okta.go
+++ b/pkg/provider/okta/okta.go
@@ -176,20 +176,27 @@ func (oc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 		body, err := ioutil.ReadAll(res.Body)
 		if err != nil {
 			return "", errors.Wrap(err, "error retrieving body from response")
-		}
-		resp = string(body)
-		re := regexp.MustCompile("var stateToken = '(.*)';")
-		match := re.FindStringSubmatch(resp)
-		if len(match) < 1 {
+		}		
+		stateToken, err := getStateTokenFromOktaPageBody(string(body))
+		if err != nil {
 			return "", errors.Wrap(err, "error retrieving saml response")
 		}
-		loginDetails.StateToken = strings.Replace(match[1], `\x2D`, "-", -1)
+		loginDetails.StateToken = stateToken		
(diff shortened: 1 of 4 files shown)


joshk0:
I can confirm this works for me and to be more precise, it fixes the case where an app level MFA policy is enabled on AWS. This change is not necessary _unless_ that is the case, and it fixes interaction with all second factors, not just TOTP.

pkg/provider/okta/okta.go
```diff
@@ -160,7 +165,26 @@ func (oc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
...
+		res, err = oc.client.Do(req)
+		if err != nil {
+			return samlAssertion, errors.Wrap(err, "error retrieving app response")
+		}
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return "", errors.Wrap(err, "error retrieving body from response")
+		}
+		resp = string(body)
+		re := regexp.MustCompile("var stateToken = '(.*)';")
```
wolfeidau: Would be good to lift this section out to a standalone function and write a small test with a couple of examples.

alsmola: See https://github.com/Versent/saml2aws/pull/369/commits/3221a0d5ac95dae02e7fb2284366b223c5f3822b

Issue #330, opened 2019-08-07
Is SAML2AWS supported Azure MFA when integrating with ADFS 4.0 (2016)


JimmyLS:
In [this](https://github.com/Versent/saml2aws/blob/master/saml2aws.go) article I found for ADFS only VIP and RSA are supported as MFA Provider. Now we are trying to enable Azure MFA as Provider and use it to enforce secondary authentication, but struggled to get it working. We have attempted both Mobile app notification or Verification Code as MFA methods. 

May I know if Azure MFA is supported? Or any special steps we missed?

Thank you.

bsx:
We ran into the same issue with Azure MFA in ADFS. I have a branch (https://github.com/bsx/saml2aws/tree/azure-mfa) where i'm currently working on integrating both app notification and verification code. If you are interessted you can test it out. Once i've cleaned up the branch i'll PR it for integration.

JimmyLS:
Oh Thank you very much. We will try it at our environment. 
Does it works fine in your environment?

bsx:
It works in our environment, I have tested it with both the verification code as well as the app notification/approval. I however reverse engineered it mostly from the HTML forms and i'm not sure if they are similar in every implementation. If you could try it out and let me know if it works for you that would be very helpful.
One word of warning: it will currently wait indefinitely if the MFA is set to app approval and the approval does not happen. I plan to add a timeout there as well. Selection of MFA method if multiple methods are available is also not implemented, it will just use whatever method is presented by default.

FernandoMiguel:
@bsx i assume this isn't merged yet?

can you merger master into your branch pls so we can build the latest?

wolfeidau:
@bsx would be good to get a PR for this so others can try it out.

Current setup would still be handy for users.

PR #380, opened 2019-11-22
refactor ADFS provider to support AzureMFA as well

bsx:
This implements both the prompt for a token as well as waiting for app confirmation (see #330). There is currently no way to actually chose the MFA type you expect, it'll just figure out what to use (mfa=Auto). There is also no timeout for waiting for app confirmation currently.


alFReD-NSH:
I tested this and it works.

wolfeidau:
Thanks for verifying this @alFReD-NSH and posting a note here it :100: helps me as I don't have access to ADFS :smile:

Issue #351, opened 2019-09-26
[AzureAD] Seems AzureAD login procedure goes alright but returns `Response did not contain a valid SAML assertion`


adamist521:
I have a problem logging with Azure AD.
From the dumped result, it seems that login procedure looks successful from the token(cookies seems related to auth which I see when I look into browser session) showing but shows `Response did not contain a valid SAML assertion`.
Haven't looked into code yet but has anyone surpassed this issue?

Looks it was different to #350 from the dumped result so I posted new Issue.

I have attached DUMP result at following. (masked token with `******`, removed JS code seemed unrelated. JSON is formatted also.)

```
$ DUMP_CONTENT=true saml2aws login --verbose
DEBU[0000] Running                                       command=login
DEBU[0000] check if Creds Exist                          command=login
DEBU[0000] Expand                                        name=/Users/xxxxxxx/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/Users/xxxxxxx/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/Users/xxxxxxx/.aws/credentials pkg=awsconfig
Using IDP Account default to access AzureAD https://account.activedirectory.windowsazure.com
DEBU[0000] Get credentials                               helper=osxkeychain serverURL="https://account.activedirectory.windowsazure.com"
DEBU[0000] Get credentials                               goMsg="The specified item could not be found in the keychain." helper=osxkeychain
To use saved password just hit enter.
? Username <*emailaddress*>
? Password ********

DEBU[0005] building provider                             command=login idpAccount="account {\n  AppID: <*AppID*>\n  URL: https://account.activedirectory.windowsazure.com\n  Username: <*emailaddress*>\n  Provider: AzureAD\n  MFA: Auto\n  SkipVerify: false\n  AmazonWebservicesURN: urn:amazon:webservices\n  SessionDuration: 3600\n  Profile: saml\n  RoleARN: \n}"
Authenticating as <*emailaddress*> ...
POST /common/login HTTP/1.1
Host: login.microsoftonline.com
User-Agent: saml2aws/1.0 (darwin amd64) Versent
Content-Length: 1375
Content-Type: application/x-www-form-urlencoded
Accept-Encoding: gzip

ctx=rQIIAZW******&flowToken=AQABA******&login=<*emailaddress*>&passwd=<*password*>
HTTP/1.1 200 OK
Cache-Control: no-cache, no-store
[...]

PR #381, opened 2019-11-24
fix aad `Response did not contain a valid SAML assertion`

paihu:
fix  https://github.com/Versent/saml2aws/issues/351


markaltmann:
I need this as well. When is it planned to be merged.
@paihu Does that fix your AAD login?

thx

Issue #382, opened 2019-11-25
Okta w/ push MFA not working in 2.20.0


choppedpork:
Hello,

After upgrading to 2.20.0 I'm no longer to successfully authenticate against Okta with a `error authenticating to IdP: Unknown document type`. The process fails at the second request to Okta (`https://xxx.okta.com/login/sessionCookieRedirect(...)`) which seems to return the login page.  My config is as follows:
```
[default]
app_id               =
url                  = https://xxx.okta.com/home/amazon_aws/0xxxx1lxxxxXXXxX123/123
username             = xxx@xxx.com
provider             = Okta
mfa                  = PUSH
skip_verify          = false
timeout              = 0
aws_urn              = urn:amazon:webservices
aws_session_duration = 7200
aws_profile          = default
resource_id          =
subdomain            =
role_arn             =
```
I can reliably reproduce this and provide more detailed logs if needed - having had a look at the release notes I would assume this has _something_ to do with #376 but that's just a guess. Downgrading to 2.19.0 fixes the issue. Let me know if there's anything else I can provide to help.

Thanks!

Anton0:
That's interesting, if you can reliably reproduce on 2.20.0 but it works with 2.19.0 then it's probably related to my change #376 however I tested using Okta PUSH and it works fine here.

Are you able to provide some logs (removing any sensitive data). Additionally can you confirm if you're getting the PUSH request from Okta - is it failing after you respond to the MFA request or before?

choppedpork:
Hey @Anton0, apologies it took me a while! After reading your comment I've realised that I'm *not* getting to the push stage in 2.20.0 so this might be something else entirely but definitely only experiencing this in 2.20.0.

Here's some `--verbose` output (edited both for sensitive data as well as some clarity):

```
$ /usr/local/bin/saml2aws --version
2.20.0
$ /usr/local/bin/saml2aws login --verbose
DEBU[0000] Running                                       command=login
DEBU[0000] check if Creds Exist                          command=login
DEBU[0000] Expand                                        name=/Users/xxx/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/Users/xxx/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/Users/xxx/.aws/credentials pkg=awsconfig
Using IDP Account default to access Okta https://x.okta.com/home/amazon_aws/asdf/272
[...]

PR #387, opened 2019-12-04
Fix Okta MFA regression issue

alsmola:
I tested this and it fixed https://github.com/Versent/saml2aws/issues/382 for me

Since the page for MFA in Okta desn't contain `form[action=\"https://signin.aws.amazon.com/saml\"]"` in `docIsFormRedirectToAWS`, the whole `if` block is skipped and we get the `Unknown document type` error.

This moves the MFA requesting code outside of that check so that it's hit if none of the other conditions are met.

### DIFF
--- a/pkg/provider/okta/okta.go
+++ b/pkg/provider/okta/okta.go
@@ -183,25 +183,6 @@ func (oc *Client) follow(ctx context.Context, req *http.Request, loginDetails *c
 			}
 			logger.WithField("type", "saml-response").WithField("saml-response", string(decodedSamlResponse)).Debug("doc detect")
 			return samlResponse, nil
-		} else {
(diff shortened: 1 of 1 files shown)


Anton0:
Thanks for doing this, I've pulled your branch and tested and this fix works.

I was going to add a test to the response to check if it was the app level MFA prompt rather than doing it as a general catch all but I think in the interest of getting things working again we should merge this fix.

alsmola:
Any updates here @Anton0? Who needs to approve for us to merge?

alsmola:
As a heads up I have a branch that supports WebAuthn once this merges: https://github.com/alsmola/saml2aws/pull/1

alsmola:
@wolfeidau thanks! Any idea when you'll push a new version? Hoping to release this to my company soon and would love to let people use the default install instructions.

PR #397, opened 2020-01-10
Add support for Keycloak 8.0.1 theme with enabled OTP

rma945:
Starting from version 8.0.0 - Keycloak was changed name of OTP token field in templates, from totp to otp, so saml2aws can't find this field. I added support for a new OTP token field.


crunchtime-ali:
My organisation needs this merged asap. Can I help somehow?

PR #420, opened 2020-02-01
fix(console) Updated issuer to use correct value

wolfeidau:


README.md
```diff
@@ -131,32 +131,63 @@ Flags:
...
 
 Commands:
   help [<command>...]
     Show help.
 
+
   configure [<flags>]
     Configure a new IDP account.
 
+        --app-id=APP-ID            OneLogin app id required for SAML assertion. (env: ONELOGIN_APP_ID)
```
onyxraven: this flags layout makes a lot more sense :)

Issue #429, opened 2020-02-11
[v2.22.1] Okta push waiting error after 1 second


paskal:
## Problem description

Okta Push verification breaks instead of waiting for my reaction after one second.

```bash
$ saml2aws login --force
Using IDP Account default to access Okta https://okta.example.com/home/amazon_aws/0ga1t1us0akS0kbyb416/272
To use saved password just hit enter.
? Username paskal@example.com
? Password

Authenticating as paskal@example.com ...
? Select which MFA option to use PUSH MFA authentication

Waiting for approval, please check your Okta Verify app ...........error authenticating to IdP: error verifying MFA: error retrieving verify response: Post https://okta.example.com/api/v1/authn/factors/opfytj55ezNSIERWE115/verify: net/http: HTTP/1.x transport connection broken: http: ContentLength=60 with Body length 0
```

## My take on the problem

By the link, there is a description of [issuing Okta Push Factor Challenge](https://developer.okta.com/docs/reference/api/factors/#verify-a-push-factor-challenge).

It states that one should POST to URL `/api/v1/users/${userId}/factors/${factorId}/verify` to obtain `transactionId`  and then GET URL `/api/v1/users/${userId}/factors/${factorId}/transactions/${transactionId}` to wait for result.

Here is the current code:
https://github.com/Versent/saml2aws/blob/e602111db872da7dd308f7ead7e70b18974642e7/pkg/provider/okta/okta.go#L373-L418

It seems to me that saml2aws currently trying to reach the POST URL every time instead of reaching the second one from the documentation with `transactionId` with GET method.

I'll prepare a PR with the fix if my theory will turn out to be right.

//cc @wolfeidau

paskal:
I'm wrong about the API used, I see that it's [Verify WebAuthn Factor ](https://developer.okta.com/docs/reference/api/authn/#verify-webauthn-factor). Also, I see that the sleep time in the following part of code is irrelevant, both with the default value and `Time.Second * 3` it's failing exactly after 8 requests, on 9th one.

https://github.com/Versent/saml2aws/blob/e602111db872da7dd308f7ead7e70b18974642e7/pkg/provider/okta/okta.go#L399-L402


Continueing investigation.

paskal:
[...]

PR #430, opened 2020-02-11
Increase waiting sleep for Okta Push to 3 seconds

paskal:
In line with `IdentifierDuoMfa`:

https://github.com/Versent/saml2aws/blob/e602111db872da7dd308f7ead7e70b18974642e7/pkg/provider/okta/okta.go#L562

I've stumbled upon a problem (#429) where Okta Push MFA getting an error after 8 successful calls to API, which is with current timeouts is 800ms. With this change in place, I'll still get an error after 8 successful calls, but it will give me 24 seconds to push "Yes, it's me" button on my phone in the Okta app instead of current ~1 second.

Details of the problem are in the linked issue.

### DIFF
--- a/pkg/provider/okta/okta.go
+++ b/pkg/provider/okta/okta.go
@@ -397,7 +397,7 @@ func verifyMfa(oc *Client, oktaOrgHost string, loginDetails *creds.LoginDetails,
 			switch gjson.Get(string(body), "factorResult").String() {
 
 			case "WAITING":
-				time.Sleep(1000)
+				time.Sleep(3 * time.Second)
 				fmt.Printf(".")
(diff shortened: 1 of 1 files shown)


paskal:
Previously users will get their verification acknowledged by the saml2aws almost immediately and will get their credentials saved after that. After this PR will be merged, the user will have to wait up to three seconds to get the credentials after they push "Yes, it's me" in the Okta app.

As a side note, this is similar to how okta.com web interface itself works for me: after I push the button in the app, I need to wait a few seconds before browser will recognize it and move to the next sceren.

paskal:
@wolfeidau?

Issue #32, opened 2017-05-04
Write diagnostic output to stderr


jghaines:
Prompts and auth details should be written to stdout.

This is especially important when using the `exec` variant as I want to pipe the output of the subcommand ignoring the saml2aws output.

wolfeidau:
Yes I need to tidy this up, I am just adding keychain support to save creds so I can sort this out.

wolfeidau:
Discovered the other day that this cause issues on windows, I will need to write a simple writer which on *nix stderr, on windows stdout.

Still searching around for existing precedence.

wolfeidau:
I have added a new --debug logger function which writes to stderr. 

Need to test this on windows in powershell cc @davidobrien1985

d3v1an7:
Hey @wolfeidau! Am putting together a wrapper for a bunch of tools, and on non zero exit status I return stderr as a hint for the user. Problem is, looks like all output (including errors) for `saml2aws` seems to still be going to stdout...

<img width="606" alt="Screen Shot 2019-09-22 at 10 42 10 am" src="https://user-images.githubusercontent.com/4087799/65380777-f24a6100-dd25-11e9-93d9-8a1cafe9c0a3.png">

Tried adding `--debug`, but looks like that has been removed. Any plans to revisit this one?

MichaelMallett:
Hi @wolfeidau I'm also keen for this, I can't see any mention of a --debug flag, and I can't seem to find anything in the commit history for one either?

valtaz:
there is a --verbose flag for "Enable verbose logging"

MichaelMallett:
Hi @valtaz, I can see the verbose flag, however from my testing it seems errors are still written to stdOut instead of stdErr, which is my issue.

PR #443, opened 2020-02-27
enable option to make login quiet

marcelobartsch:
This change allow to hidden the 'credentials are not expired skipping' message.

This is useful for me as I use the saml2aws in my oh-my-zsh setup everytime before I run a command, and if the credentials are expired, they will get updated. and the message above is a bit annoying to see in the prompt all the times.

### DIFF
--- a/cmd/saml2aws/main.go
+++ b/cmd/saml2aws/main.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/alecthomas/kingpin"
 	"github.com/sirupsen/logrus"
+
 	"github.com/versent/saml2aws/cmd/saml2aws/commands"
 	"github.com/versent/saml2aws/pkg/flags"
 )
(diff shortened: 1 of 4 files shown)


RichiCoder1:
:+1: to this

cmd/saml2aws/main.go
```diff
@@ -86,6 +87,7 @@ func main() {
 	cmdLogin.Flag("client-id", "OneLogin client id, used to generate API access token. (env: ONELOGIN_CLIENT_ID)").Envar("ONELOGIN_CLIENT_ID").StringVar(&commonFlags.ClientID)
 	cmdLogin.Flag("client-secret", "OneLogin client secret, used to generate API access token. (env: ONELOGIN_CLIENT_SECRET)").Envar("ONELOGIN_CLIENT_SECRET").StringVar(&commonFlags.ClientSecret)
 	cmdLogin.Flag("force", "Refresh credentials even if not expired.").BoolVar(&loginFlags.Force)
+	cmdLogin.Flag("quiet", "Supress credentials not expired message").BoolVar(&loginFlags.Quiet)
```
RichiCoder1: May this should just be "Suppresses information messages" as it'd be neat if this could silence the `login` output if you `--skip-prompt`.

marcelobartsch: Fully Agree on the message.
On the --skip-prompt. when login is required, I'm not sure is a good idea, at least I would leave the 'Waiting for approval, please check your XXXX Verify app' message visible.

marcelobartsch: Changes done, please my apologies if the way is not the best, I'm not a programmer just a DevOps/SRE/SysAdmin :)
hopefully it is good enough.

wolfeidau:
I really like the idea of this, I am just not super keen on the implementation.

 It is linked to #32 which needs to do different things if we are on Linux and Windows, I really need something which finds all fmt calls and replace them with a global writer which has a quite mode.

Thoughts on this?

oshaughnessy:
@marcelobartsch, I did something similar in #605. Does that take care of your need, too?

mapkon (changes requested):
@marcelobartsch Is this still required? If yes, can you update to achieve parity with master with regards to conflicts and test coverage?

PR #445, opened 2020-02-28
Add support for new Google Login Page Formats

### LINKED ISSUE #203: Error authenticating using google
After accepting yes on google prompt during sam2aws login
Getting exception
```
error authenticating to IdP: page is missing saml assertion
```

Did configure
```
account {
  URL: https://accounts.google.com/
  Username: rohit.verma@xxx.xx
  Provider: GoogleApps
  MFA: Auto
  SkipVerify: false
  AmazonWebservicesURN: urn:amazon:webservices
  SessionDuration: 3600
  Profile: default
}
```
with .aws/config file as
```
[default]
region = us-west-2
google_config.ask_role = False
google_config.keyring = False
google_config.duration = 43100
google_config.google_idp_id = C000000
google_config.role_arn = arn:aws:iam::0000000000000:role/root-sysadmin-delegate
google_config.google_sp_id = 0000000000
google_config.u2f_disabled = True
google_config.google_username = rohit.verma@xxx
```

### LINKED ISSUE #444: GoogleApps: unsupported protocol scheme
Hello,

We use saml2aws on a daily basis and have suddenly started running into this error when logging in (about 1 in every 5 logins):
```Post ?hl=en&loc=US: unsupported protocol scheme ""```

After running in debug mode it looks like the `passwordURL` variable in the below url is intermittently not being set: https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L59

debug output:
```time="2020-02-27T13:41:34-05:00" level=debug msg="loginURL: " provider=googleapps```

After lots of digging by our dev team we saw that when a failure occurs, saml2aws is sent to the login page asking for a username. When it works, its sent to the login page asking for a password.

The function in googleapps.go `extractInputsByFormQuery` seems to be looking for an html element called `gaia_loginform` which looks like the below.
On failure:
```<form novalidate method="post" action="https://accounts.google.com/signin/v1/lookup" id="gaia_loginform">```
on success::
```<form novalidate method="post" action="https://accounts.google.com/signin/challenge/sl/password" id="gaia_loginform">```

We were able to work around the issue temporarily by adding an if on line 465 of `googleapps.go`:
```diff
diff --git a/pkg/provider/googleapps/googleapps.go b/pkg/provider/googleapps/googleapps.go
index 2af5a87..1017ec0 100644
--- a/pkg/provider/googleapps/googleapps.go
+++ b/pkg/provider/googleapps/googleapps.go
@@ -463,6 +463,10 @@ func extractInputsByFormQuery(doc *goquery.Document, formQuery string) (url.Valu
                actionURL = action
        })
 
+       if actionURL == ""{
+               actionURL = "https://accounts.google.com/signin/challenge/sl/password"
+       }
+
        query = fmt.Sprintf("form%s", formQuery)
 
        // extract form data to passthrough
```

But that work around is obviously not the best implementation.

nathandines:
Fixes #203, fixes #444

Google has been changing a number of their login pages recently. As a
result of that, their CAPTCHA page doesn't behave exactly as before, nor
do some of the login pages themselves.

This change is designed to be backwards compatible with the existing
Google login and captcha pages, as it appears that the changes are not
completely propogated through their systems at this time. This has made the change a bit verbose, but by necessity (this whole provider could probably use a refactor at some point).

This commit contains a fair number of changes, among which are:

- Support for generating absolute URLs from relative paths
    - Some of the fields being references in `src` and `action`
    attributes are now relative, so we need to expand them
- Support for logging into a final password page after the CAPTCHA is
complete, as this behaviour now seems different
- Support for different fields which are now used by the Google login pages

I've had a few people test this, but would be happy for the broader community to give it a go as well. I believe that the GoogleApps provider will be broken for the most part until this is merged.

### DIFF
--- a/pkg/provider/googleapps/googleapps.go
+++ b/pkg/provider/googleapps/googleapps.go
@@ -66,37 +66,74 @@ func (kc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 		return "", errors.Wrap(err, "error loading challenge page")
 	}
 
-	captchaFound := responseDoc.Find("#logincaptcha")
+	captchaInputIds := []string{
+		"logincaptcha",
+		"identifier-captcha-input",
+	}
+
+	var captchaFound *goquery.Selection
+	var captchaInputId string
+
+	for _, v := range captchaInputIds {
+		captchaFound = responseDoc.Find(fmt.Sprintf("#%s", v))
+		if captchaFound != nil && captchaFound.Length() > 0 {
+			captchaInputId = v
+			break
+		}
+	}
 
 	for captchaFound != nil && captchaFound.Length() > 0 {
-
 		captchaImgDiv := responseDoc.Find(".captcha-img")
-		captchaPictureURL, found := goquery.NewDocumentFromNode(captchaImgDiv.Children().Nodes[0]).Attr("src")
+		captchaPictureSrc, found := goquery.NewDocumentFromNode(captchaImgDiv.Children().Nodes[0]).Attr("src")
 
 		if !found {
 			return "", errors.New("captcha image not found but requested")
+		}
+
+		captchaPictureURL, err := generateFullURLIfRelative(captchaPictureSrc, passwordURL)
+		if err != nil {
+			return "", errors.Wrap(err, "error generating captcha image URL")
 		}
 
 		fmt.Println("Open this link in a browser:\n", captchaPictureURL)
 
 		captcha := prompter.String("Captcha", "")
 
-		captchaForm, captchaURL, err := extractInputsByFormID(responseDoc, "gaia_loginform")
+		captchaForm, captchaURL, err := extractInputsByFormID(responseDoc, "gaia_loginform", "challenge")
 		if err != nil {
 			return "", errors.Wrap(err, "error extracting captcha")
 		}
 
 		logger.Debugf("captchaURL: %s", captchaURL)
 
-		captchaForm.Set("Passwd", loginDetails.Password)
-		captchaForm.Set("logincaptcha", captcha)
+		_, captchaV1 := captchaForm["Passwd"]
+		if captchaV1 {
+			captchaForm.Set("Passwd", loginDetails.Password)
+		}
+		captchaForm.Set(captchaInputId, captcha)
 
 		responseDoc, err = kc.loadChallengePage(captchaURL+"?hl=en&loc=US", captchaURL, captchaForm, loginDetails)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mtse-aurora:
The fix works for me!

lifeofguenter:
Seems like the login is going to fail gradually worse - so would be cool if we could get this merged in and released, as we also need to update brew :)

wdalmut:
:+1:

pkg/provider/googleapps/googleapps.go
```diff
@@ -51,52 +51,102 @@ func (kc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
...
 
 	logger.Debugf("loginURL: %s", passwordURL)
 
 	authForm.Set("Passwd", loginDetails.Password)
-	authForm.Set("rawidentifier", loginDetails.Username)
 
-	responseDoc, err := kc.loadChallengePage(passwordURL+"?hl=en&loc=US", authURL, authForm, loginDetails)
+	referingURL := passwordURL
+
+	if _, rawIdPresent := passwordForm["rawidentifier"]; rawIdPresent {
```
andriiz42: I suggest using the variable name `rawIDPresent`.
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

wdalmut:
We have just a problem when you have multiple accounts to choose, the default selection is not immediately valid and you have to move down and then up to select the first proposed account.

aknysh:
thank you for fixing it @nathandines ! it affected as as well.
Can you please cut a new release @wolfeidau ?

wdalmut:
Today when i try the new master i get this error

```sh
Authenticating as myuser@domain.tld ...
error authenticating to IdP: page is missing saml assertion
```

Lemmons:
> Today when i try the new master i get this error
> 
> ```shell
> Authenticating as myuser@domain.tld ...
> error authenticating to IdP: page is missing saml assertion
> ```

The [current release](https://github.com/Versent/saml2aws/releases/tag/v2.23.0) is confirmed to be working though.  Definitely do not release as is.

nathandines:
That’s really odd... I suspect that Google have been changing their login pages further. It’s also possible that they rolled back their changes. This patch was intended to be backwards-compatible, but I wasn’t able to hit the previous version of the login page, so couldn’t test that and might have missed something.

I won’t have a chance to jump on this any time soon, but hopefully it’s just a minor change that somebody else can take a look at.

It’s also possible that Google aren’t finished making changes, so further near-term patches might also not last.

nyodas:
I made a small patch that seems to fix it for me in #449

PR #449, opened 2020-03-02
[provider][googleapps] - Add bgresponse=js_disabled

nyodas:
Following #445 and [this issue in the pr](https://github.com/Versent/saml2aws/pull/445#issuecomment-593485448)

The login via GoogleApps was not working for me.
The Authentication page was consistently returning 400 instead
of the 302 Redirect that is followed by a 200 OK
Adding the `bgresponse` to the formData fixed it.


wdalmut:
:+1: i can signup this morning with this patch using google provider

PR #450, opened 2020-03-02
Allow for govcloud saml redirect

nyodas:
Currently the SAML use the default commercial partition for its calls.
This has the side-effect of preventing govcloud from working.
I used the fastest way to go there.

But we could parse the XML to use the saml2p:Response->Destination to use the correct target without having to rely on a `grep` of the content

And the region/partition could also be inferred in a similar way.

### DIFF
--- a/aws_account.go
+++ b/aws_account.go
@@ -28,6 +28,10 @@ func ParseAWSAccounts(samlAssertion string) ([]*AWSAccount, error) {
 		fmt.Println("trying to login AWS China")
 		awsURL = "https://signin.amazonaws.cn/saml"
 	}
+	if strings.Contains(string(decSamlAssertion), "signin.amazonaws-us-gov.com") {
+		fmt.Println("trying to login AWS Govcloud")
+		awsURL = "https://signin.amazonaws-us-gov.com/saml"
+	}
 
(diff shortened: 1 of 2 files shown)


missingcharacter:
FYI, I created a PR that allows for this without hardcoded URLs https://github.com/Versent/saml2aws/pull/452

nyodas:
> FYI, I created a PR that allows for this without hardcoded URLs #452

Unfortunately it does not appear to work.
Will look as to why later on, it seems to be a better way to do it , than the hardcoding of url .

missingcharacter:
@nyodas I currently use it with gov-cloud, you need to specify the region:

```toml
[example]
app_id               =
url                  = https://accounts.google.com/o/saml2/initsso?idpid=<yours>j&spid=<yours>&forceauthn=false
username             = user@gsuite.com
provider             = GoogleApps
mfa                  = Auto
skip_verify          = false
timeout              = 0
aws_urn              = urn:amazon:webservices
aws_session_duration = 28800
aws_profile          = example
resource_id          =
subdomain            =
role_arn             = role:arn
region               = us-gov-west-1
```

nyodas:
Closing this in favor of #452

PR #456, opened 2020-03-08
Feat/yubikey support

edyesed:
Hi,

Yubikeys have a slightly deviant login flow to normal second factors. 
I've added support for the way that their login flows work via this PR. 

This is working for me in our oktapreview account. I'm going to see if I can get some additional internal testers via other org yubikey users. 

I'm not really a go-ologist, so if this is not ideomatic or whatever, feel free to modify.  I tried to adopt the existing approaches.


edyesed:
hi @wolfeidau , it isn't clear how to get a reviewer for this PR. Can you or somebody else guide me through the process? My place of employment would very much like to have this feature ( even if it isn't filled by this particular PR ).

paskal:
@edyesed thanks a lot, works flawlessly!

PR #458, opened 2020-03-09
Added support for DUO authorized network bypass

ckrusch:
Added support to the shibboleth provider for DUO authorized network bypass.

Compared saml2aws debug output for non-bypassed and bypassed cases.  Added a few lines of code to handle the bypass case.

Attached credential sanitized capture of saml2aws output from a network authorized for Duo bypass. The required token is returned much earlier in the transaction for authorized networks at line 160 in the js_cookie hidden input field.
[samlDuoNetBypass.txt](https://github.com/Versent/saml2aws/files/4307262/samlDuoNetBypass.txt)


ckrusch:
Tested and validated bypass fix working correctly on

Windows 10 Version 1909
MacOS Catalina 10.15.4

PR #463, opened 2020-03-12
Feature: http retry

ilya-v-trofimov:
Adds an ability to configure retries for http calls


pkg/provider/netiq/netiq.go
```diff
@@ -88,6 +88,26 @@ func (nc *Client) follow(req *http.Request, loginDetails *creds.LoginDetails) (s
 	}
 }
 
+//func (nc *Client) Do(req *http.Request) (*http.Response, error) {
```
chengchengmu: do we still need this commented code ?

chengchengmu:
Very nice feature ! Well done
LGTM

pkg/provider/netiq/netiq.go
ilya-v-trofimov: Good catch. Removing now

PR #466, opened 2020-03-16
Add support for Duo Access Gateway

multimac:
Added support for the Duo Access Gateway as a SAML IdP.

Built this by watching the requests made when authenticating to AWS via SAML directly from the Duo Access Gateway and recreating those requests in Go.


mapkon (changes requested):
@multimac Do you still need this? If yes, can you fix the conflicts so that I can review this for a merge?

PR #479, opened 2020-04-10
Make `config` flag global

gmiroshnykov:
I was surprised to discover that `--config` flag and `SAML2AWS_CONFIGFILE` env var was only respected by `configure` subcommand, so I fixed it by making the flag global.

Please let me know if I misunderstood or missed something.

Thanks!


laggyluke:
Hey @wolfeidau, any chance we could merge this?
Or is there anything I should change?

Issue #471, opened 2020-03-19
Typo in fishTmpl


hacker65536:
https://github.com/Versent/saml2aws/blob/96998410abc0c1b0454b47dee3b703b104480575/cmd/saml2aws/commands/script.go#L27-L28

Error Log
--
```console
>  saml2aws script --shell=fish
--snip--
set -gx AWS_CREDENTIAL_EXPIRATION 'error generating template: template: envvar_script:6:38: executing "envvar_script" at <.Expires>: Expires has arguments but cannot be invoked as function
```

Environment
--
```console
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.3
BuildVersion:	19D76
$ fish --version
fish, version 3.1.0
$ saml2aws --version
2.24.0
```

PR #484, opened 2020-04-15
fix script's template for powershell

pseidel-kcf:
same as https://github.com/Versent/saml2aws/issues/471 but for powershell

`saml2aws.exe script --shell=powershell`
// snip
`$env:AWS_CREDENTIAL_EXPIRATION='error generating template: template: envvar_script:6:35: executing "envvar_script" at <.Expires>: Expires has arguments but cannot be invoked as function`


wolfeidau:
👍 thanks

PR #490, opened 2020-04-22
add Privileged authentication option in NetIQ

chengchengmu:
Privileged authentication is skipping MFA whereas default authentation uses MFA, hence it's implemented as a MFA option
Privileged authentication uses a different login path than the default one.


wolfeidau:
Code, documentation and tests in a PR 😂 👍

PR #502, opened 2020-05-07
feat(logging) All logs to stderr on *nix and stdout on windows

### LINKED ISSUE #32: Write diagnostic output to stderr

wolfeidau:
* Removed spinner due to output issues

Fixes  #32


d3v1an7:
Awesome! Thanks, will check this out.

PR #503, opened 2020-05-11
Correct end index when extracting JSON from login response

nenad:
Semicolon can show up as a character in the response and not signify end of statement.

For example, on my company's Azure login I get this in the response:
```
$Config={... "Illustration":"https://aadcdn.msauthimages.net/c1c6b6c8-ujbrmqystroz2zskhz5wa4ahmdyaiqbvcjcapkylqvg/logintenantbranding/0/illustration?ts=636978528194440357","BoilerPlateText":"Please use your Azure username to sign in. \n\nAttention: Your Azure password is different from your Gmail password. Azure and Gmail are different user accounts with each their own credentials (it just so happens that Azure username and Gmail have the same format).\n\nA strong password complies with the following:\n- A minimum of 8 characters \n- Requires three out of four of the following:\nLowercase characters.\nUppercase characters.\nNumbers (0-9).\nSymbols (@ # $ % ^ &amp; * - _ ! + = [ ] { } | \\ : ‘ , . ? / ` ~ &quot; ( ) ;)","UserIdLabel":"Please use your Azure username to sign in ","KeepMeSignedInDisabled":true,"UseTransparentLightBox":false}] ...};
```

Using a new line fixes this issue as it matches literal newline character and not an escaped one, as seen in the example above.

### DIFF
--- a/pkg/provider/aad/aad.go
+++ b/pkg/provider/aad/aad.go
@@ -699,7 +699,7 @@ func (ac *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	var loginPasswordJson string
 	if strings.Contains(resBodyStr, "$Config") {
 		startIndex := strings.Index(resBodyStr, "$Config=") + 8
-		endIndex := startIndex + strings.Index(resBodyStr[startIndex:], ";")
+		endIndex := startIndex + strings.Index(resBodyStr[startIndex:], "\n") - 1
 		loginPasswordJson = resBodyStr[startIndex:endIndex]
 	}
 	var loginPasswordResp passwordLoginResponse


pkg/provider/aad/aad.go
```diff
@@ -699,7 +699,7 @@ func (ac *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	var loginPasswordJson string
 	if strings.Contains(resBodyStr, "$Config") {
 		startIndex := strings.Index(resBodyStr, "$Config=") + 8
-		endIndex := startIndex + strings.Index(resBodyStr[startIndex:], ";")
+		endIndex := startIndex + strings.Index(resBodyStr[startIndex:], "\n") - 1
```
kenchan0130, line 702: This process gets the Object assigned to the `$Config` variable of JavaScript as a JSON string.

The end statement of JavaScript is `;`.

However, if only `;` is used, the corresponding character may be included in the JSON string.
If you want to parse more strictly, I think that it is better to specify `};`.

nenad: That was my initial approach actually. However, if you take a look at the returned HTML, there is a clear separation of these variables per newline.
The reason I've specified `\n` as an end index is simple: It's not possible to put `\n` character in the returned JS code because it will have to be escaped. This reason, combined with the fact that each statement is on a new line seemed to be a good start/end indicator.

kenchan0130: Depending on the page, it may return a response with a javascript object containing a newline as follows.

```js
$Config = {
  "Illustration": "https://aadcdn.msauthimages.net/c1c6b6c8-ujbrmqystroz2zskhz5wa4ahmdyaiqbvcjcapkylqvg/logintenantbranding/0/illustration?ts=636978528194440357",
  "BoilerPlateText": "Please use your Azure username to sign in. \n\nAttention: Your Azure password is different from your Gmail password. Azure and Gmail are different user accounts with each their own credentials (it just so happens that Azure username and Gmail have the same format).\n\nA strong password complies with the following:\n- A minimum of 8 characters \n- Requires three out of four of the following:\nLowercase characters.\nUppercase characters.\nNumbers (0-9).\nSymbols (@ # $ % ^ &amp; * - _ ! + = [ ] { } | \\ : ‘ , . ? / `~ & quot; ();)",
  "UserIdLabel ":"Please use your Azure username to sign in",
  "KeepMeSignedInDisabled": true,
  "UseTransparentLightBox": false
};
```

In this case, I think it's best to avoid using `\n` because it won't parse correctly.

nenad: You can take a look at the example response returned in my PR description. It contains `\n` in the returned response and is indeed parsed correctly because we are not matching an escaped character but the newline character. If we were matching against `\n` in backticks and not `"\n"` then yes, we would have the parsing error as that matches verbatim.

Issue #527, opened 2020-07-22
Okta + Duo MFA broken due to DuoSecurity API change


himanshu-neema:
#### Failures observed on release v2.14.0

On successful authentication with Okta & successful approval of Duo push notification, `saml2aws` fails with following error:

```
➜  ~ saml2aws login --duo-mfa-option="Duo Push"
Using IDP Account default to access Okta <redacted-IDP-url>
To use saved password just hit enter.
? Username <redacted-email>
? Password **************
Authenticating as <redacted-email> ...
Pushed a login request to your device...
Success. Logging you in...
Response did not contain a valid SAML assertion
Please check your username and password is correct
```

#### Failures. observed o n release v2.26.2 

On latest released version, when user approve Duo Push notification `saml2aws` prompts again for Duo Push,  and hence goes into infinite loop

```
➜  saml2aws git:(d947726) DBUS_SESSION_BUS_ADDRESS=/dev/null $GOPATH/bin/saml2aws login --force
Using IDP Account default to access Okta <redacted-IDP-url>
To use saved password just hit enter.
? Username <redacted-email>i
? Password *************

Authenticating as <redacted-email> ...
? Select a DUO MFA Option Duo Push
Pushed a login request to your device...
Success. Logging you in...
? Select a DUO MFA Option Duo Push
Pushed a login request to your device...
Success. Logging you in...
? Select a DUO MFA Option  [Use arrows to move, type to filter]
> Duo Push
  Passcode
```
 Note about `DBUG_SESSION_BUS_ADDRESS=/dev/null`: without this env var, `saml2aws` appears to be hung. 

### Response from Duo Security support

They rolled out the change yesterday for our account, so this may not be affecting everyone. Following is the screenshot of response:

<img width="500" alt="Screen Shot 2020-07-22 at 1 07 15 PM" src="https://user-images.githubusercontent.com/64927161/88222811-5b875300-cc1b-11ea-939c-b97383c2d42f.png">

### Desired outcome for this issue:
1. Need to fix latest code for the Duo Security API change, and still be able to handle old API behavior.
2. Backport fix to version 2.14.0 - since my team uses this version. Or should we fix the latest and upgrade to latest?

himanshu-neema:
Note related to `DBUG_SESSION_BUS_ADDRESS=/dev/null` there's an open issue https://github.com/Versent/saml2aws/issues/441

AaronAtDuo:
[...]

PR #532, opened 2020-08-04
Issue 527: Duo MFA: pass new SID to subsequent API call to DuoSecurity (with backward compatibility)

himanshu-neema:
Fixes issue: https://github.com/Versent/saml2aws/issues/527

#### Fix for Duo Security API change: 
- Store the new SID obtained in response from  `POST /frame/status` ( `duoSubmitURL` in code )
- Pass new SID in subsequent API call to `POST /frame/status/<TxID>`  ( `duoResultURL` in code )

#### Bug fixes (this was causing infinite loop for Duo Push MFA failures) :
-  Certain code paths in `verifyMfa` returns `nil` error when in fact it should return an error: `error.Wrap( )` replaced with `error.New( )` : because [wrapping `nil` `err` returns `nil`](https://godoc.org/github.com/pkg/errors#Wrap)

#### Update 18th Aug 2020
We noticed API reverted back: hence made the change backward compatible in https://github.com/Versent/saml2aws/pull/532/commits/6200cf3db26d7eb5c54e37098e43a0d884b9f292


himanshu-neema:
Hi @wolfeidau - Please review this PR when you get a chance and let me know what you think.

himanshu-neema:
Thank you!

PR #545, opened 2020-08-27
Add an option to specify the expected sign-in destination

tjamet:
The support multiple redirects and intermediate SAML relays has required to know the final AWS sign-in URL and detect it in the redirection loops. This means all the relays addresses needs to be known in advance, and support for new ones cannot be supported at all.

This change allows, while keeping a sane default value referencing all known AWS saml relays, to support any other one in case some users faces specific cases, or even, wants to use the code to authenticate to any other service provider.

### DIFF
--- a/pkg/cfg/cfg.go
+++ b/pkg/cfg/cfg.go
@@ -30,22 +30,23 @@ const (
 
 // IDPAccount saml IDP account
 type IDPAccount struct {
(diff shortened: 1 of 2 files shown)


tjamet:
Hi! any news on this?

mapkon (changes requested):
@tjamet  Is this still required? If yes, can you update to achieve parity with master with regards to conflicts and test coverage?

Issue #551, opened 2020-09-08
Link to instructions on failure is broken


iainelder:
When the login fails I see this error:

```text
Response did not contain a valid SAML assertion
Please check your username and password is correct
To see the output follow the instructions in https://github.com/versent/saml2aws/v2#debugging-issues-with-idps
```

When I visit the link I get a 404 error:

![imagen](https://user-images.githubusercontent.com/5024311/92442683-97b54980-f1b0-11ea-9405-9c1254ed9887.png)

Does it mean to link here?

https://github.com/Versent/saml2aws#debugging-issues-with-idps

![imagen](https://user-images.githubusercontent.com/5024311/92442707-a3087500-f1b0-11ea-8396-8b10e2a4501b.png)

PR #553, opened 2020-09-09
Updating broken URL for IDP Issues

jack1902:
related to:
resolves: #551

## Background

When errors occur with Authentication, the message below is displayed:
```
Response did not contain a valid SAML assertion
Please check your username and password is correct
To see the output follow the instructions in https://github.com/versent/saml2aws/v2#debugging-issues-with-idps
```

This URL, resulted in a 404 which should now be fixed

## Changes

* `https://github.com/versent/saml2aws/v2#debugging-issues-with-idps` -> `https://github.com/versent/saml2aws#debugging-issues-with-idps`

## Testing

Built the binary locally and the link now works :D


wolfeidau:
thanks for that!

PR #558, opened 2020-09-16
Add flag to fetch credentials for all roles for a given profile

goelankitt:
I intend to fetch **credentials** for all available roles for a given profile. In this PR, this is accomplished by adding a new option `--all`. So when  the user runs `saml2aws login --all`, the role selection prompt is bypassed and credentials for every available role are fetched and added to the credentials file. The profile names in the credential file in this case are autogenerated as saml1, saml2 etc. I think this is something that can be improved upon. However, please let me know if this is an acceptable approach.

### DIFF
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -4,6 +4,7 @@ import (
 	b64 "encoding/base64"
 	"log"
 	"os"
+	"strconv"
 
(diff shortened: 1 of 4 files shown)


wolfeidau:
@goelankitt have you tried the list-roles sub command?

goelankitt:
@wolfeidau Yes, I just did. It only lists the roles and does not fetch the credentials.

Issue #559, opened 2020-09-16
list-roles should print to stdout and not stderr


grosser:
2.22.1 printed the list of accounts to stdout, which allowed to pipe it to a file and still show the user some output
2.27.0 writes everything to stderr, so now I cannot separate them easily and things like `Success. Logging you in...` ends up in the output

please make the output go to stdout again

LorisCazalon:
hello, as a workaround in our scripts that get saml2aws, i propose to also get the content of the stderr : https://github.com/zendesk/dotfiles_n_scripts/pull/81. 
We should revert the PR once this is fixed

grosser:
Bug was added in https://github.com/Versent/saml2aws/pull/502

PR #560, opened 2020-09-18
The roles from list-roles are output to stderr not stdout which is better for scripting

grosser:
…eed to go to stdout

fixes https://github.com/Versent/saml2aws/issues/559
introduced by https://github.com/Versent/saml2aws/pull/502

/cc ppl involved in the original issue/PR and new issue
@wolfeidau  @LorisCazalon  @d3v1an7 @jghaines @MichaelMallett @valtaz

### DIFF
--- a/cmd/saml2aws/commands/list_roles.go
+++ b/cmd/saml2aws/commands/list_roles.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	b64 "encoding/base64"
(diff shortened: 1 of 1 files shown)


wolfeidau:
@grosser thanks for the fix will release today Melbourne/Australia time

grosser:
thx!

PR #562, opened 2020-09-25
Add support for Defender MFA.

dgrenner:
Hi!

This patch adds support for Defender MFA (https://www.oneidentity.com/products/defender/) to the ADFS provider.


dgrenner:
MFA sign-on works, but some settings from profile (role and session-duration) disappear.

wolfeidau:
Sorry it took so long to merge. 🎉

PR #565, opened 2020-09-25
Support all session options (including region) when checking the token.

stensonb:
This ensures token verification works in govcloud.  Otherwise, when tokens are verified, the wrong region may be used.


stensonb:
Without this, this blocks validating tokens via Okta -> AWS STS saml workflow for govcloud accounts.

(Although, tokens generated ARE valid, they just need to point to the correct AWS_REGION to verify).

mapkon (changes requested):
@stensonb Do you still need this? If yes, can we update it from main and plan to merge it? If not, I am inclined to close it.

Issue #566, opened 2020-09-29
Login fails if Keycloak SAML response page contains input fields without name attribute


fredriv:
The loop at https://github.com/Versent/saml2aws/blob/master/pkg/provider/keycloak/keycloak.go#L75-L87 looks for the SAML response in the form, but fails on line 78 if there are preceding input fields on the page that lack a `name` attribute.

PR #567, opened 2020-09-29
Fix keycloak login when SAML response page contains input fields without 'name' attribute

fredriv:
This fixes https://github.com/Versent/saml2aws/issues/566


wolfeidau:
Thanks for the PR, also tests are super helpful!

PR #585, opened 2020-12-03
Run all compiles with GHA

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>


gliptak:
@wolfeidau please review

The ```Required``` targets would need to be updated

Issue #419, opened 2020-01-31
Support hardware yubikey for okta provider with MFA


Richard-outreach:
Just tried using this for okta based login with a yubkey for MFA and got a `unsupported mfa provider` response. Is it possible to support `mfaIdentifer="YUBICO TOKEN:HARDWARE"` by chance?

wolfeidau:
To be honest I haven't tried using a yubikey with okta, there has been some work done in the google provider to support this device https://github.com/Versent/saml2aws/pull/398. 

Once that settles in I was planning to see if i could reuse it on other providers or come up with some common construct for these types of devices.

Cheers.

Richard-outreach:
Solid. Sounds good. Happy to test anything out that we can get rolling.

spoonerd:
+1 to this as we are running into the same issue as we use Yubikeys. Happy to do any testing required.

dishbreak:
I've got an engineer with a Yubikey as well. Haven't contributed yet, but I'm sure I'd be able to help get testing done.

edyesed:
I believe I solved this with #456

rsclarke-vgw:
Using Yubikey 5c with Okta + MFA setting `mfa = FIDO` in `~/.saml2aws` configuration has been working for me.

ylorenzati:
hello, just to let you know. It works very well on my MacBook.
But my coworkers are unable to make it work on Linux (specifically ArchLinux).

mattzuba:
@ylorenzati, I ran into this same issue with Linux, appears to be an upstream limitation: marshallbrekka/go-u2fhost#1

edyesed:
i don't have a linux machine with which to test/bring a different implementation.

kunickiaj:
Using Okta with Webauthn (FIDO) and Yubikeys. Tried but get this error:

```
error authenticating to IdP: error verifying MFA: The provided key handle is not present on the device, or was created with a different application parameter.
```

I am wondering if this is because there is a FIDO PIN set on the key. I am wondering if something needs to be added to this tool similar to how it was necessary in `gimme-aws-creds` (https://github.com/Nike-Inc/gimme-aws-creds/pull/243)

edyesed:
yeah, what `gimmie-aws-creds` is doing with the yubikey and what this PR is doing with the yubikey are quite different. Part of the difficulty with accommodating these devices are the many variants and configurations they can be put into. 

[...]

Issue #56, opened 2017-08-23
no buildable Go source files osxkeychain


debojitkakoti:
How to reproduce.
OS: Ubuntu 16.04.1 LTS
How: initiate make command.
Output: 
```
1 errors occurred:
--> darwin/amd64 error: exit status 1
Stderr: go build github.com/versent/saml2aws/helper/osxkeychain: no buildable Go source files in /home/debojit/gocode/src/github.com/versent/saml2aws/helper/osxkeychain

Makefile:29: recipe for target 'compile' failed
make: *** [compile] Error 1

```

wolfeidau:
Apologies, if you want to build on Linux use

```
make install
```

I need to update the Makefile to disable cross compile outside of OSX, I have started using `cgo` for integration with OSX keychain.

wolfeidau:
I am always willing to accept PRs to refactor this.

tomislacker:
_At the risk of sounding naive..._

@wolfeidau simply a clone and attempt at `make install` is not successful.

```
$ make install 
go install ./cmd/saml2aws
cmd/saml2aws/main.go:8:2: cannot find package "github.com/alecthomas/kingpin" in any of:
	/usr/lib/go/src/github.com/alecthomas/kingpin (from $GOROOT)
	/home/user/go/src/github.com/alecthomas/kingpin (from $GOPATH)
cmd/saml2aws/main.go:9:2: cannot find package "github.com/versent/saml2aws/cmd/saml2aws/commands" in any of:
	/usr/lib/go/src/github.com/versent/saml2aws/cmd/saml2aws/commands (from $GOROOT)
	/home/user/go/src/github.com/
```

```
 $ [ -z "$GOPATH" ] && echo "GOPATH is empty" || echo "GOPATH='${GOPATH}'"
GOPATH is empty
 $ [ -z "$GOROOT" ] && echo "GOROOT is empty" || echo "GOROOT='${GOROOT}'"
GOROOT is empty
```

If some sage words of advice could be offered to at least provide direction, I'd be willing to figure out the linux implications as that's what I predominately use. Then I can contribute a PR back to document what's all necessary.

wolfeidau:
Just run `make test` once before running `make install` to install local dev dependencies.

This is something I need to tidy up, I have built the classic "works for me" build scripts, your feedback is a valuable sign in need to tidy this up.

debojitkakoti:
It was trying to compile for all the arch when I do make install.
[...]

PR #592, opened 2021-01-15
Support U2F/WebAuthn keys USB detection in Linux

ocraviotto:
This PR updates marshallbrekka/go-u2fhost, imported by the Okta and Google providers to support U2F/WebAuthn USB keys as 2FA devices.

The update should help fix the problems reported in https://github.com/Versent/saml2aws/issues/419 as they had their origin in marshallbrekka/go-u2fhost not able to properly detect these devices in Linux. That was fixed upstream with https://github.com/marshallbrekka/go-u2fhost/pull/10 which also explains what other changes were done in the hid libraries used by go-u2fhost to make the detection work in Linux too.

Along the library change and other mod updates, this commit includes a change to the makefile to support the required tag to use
hidraw instead of libusb when compiling linux (and so really apply the fix when compiling) as well as related changes to make it easier to build/install saml2aws in different platforms, including OS detection to prevent macOS builds from building in non-Darwin environments, which should tackle https://github.com/Versent/saml2aws/issues/56.

Finally, I've also included some updates in the README to clarify the changes in the build process and the additional per OS targets.


wolfeidau:
This is a fantastic idea but i am really not keen on adding any CGO dependent stuff as it is already killing me.

I would love to figure out a way to put this in a plugin or hook which people can install separately.

fieldju:
Would love to get this merged, had to fork the fork, clone, and compile it myself :man_shrugging:.

ocraviotto:
> This is a fantastic idea but i am really not keen on adding any CGO dependent stuff as it is already killing me.

@wolfeidau Your project is already using it. I just updated the go-u2fhost ref so as to pick up the change in the underlying library used by it, so a dependency of your dependency.

Under the hood, go-u2fhost was using github.com/karalabe/hid, I just made a contribution (referenced above) to change that to github.com/bearsh/hid, which is a fork that updated karalabe's libusb and hidapi code base with upstream libusb/libusb and
libusb/hidapi, which among other things incorporates the changes in libusb/hidapi to support Usage Page and Usage on Linux with hidraw (see https://github.com/libusb/hidapi/pull/139/files) when packages are built with "-tags=hidraw", while maintaining the old behaviour when built without it. I also contributed with a fix to the fork.

In brief, this PR only adds the "-tags=hidraw" tag on Linux builds to support Usage Page and Usage on Linux with hidraw.
The rest of the changes should address other building issues but it do not include additional CGO dependencies that were not already there in previous versions.

If you want me to update my branch with master to reconsider, I'll be happy to. Let me know.

smlx:
Hi, thanks for working on this! I found this PR and rebased it onto `master` to get multiple yubikey support in Okta. The nice thing is that `master` now has the updated go-u2fhost module version so this PR can be trimmed down to only README and Makefile changes :tada: 

In any case, I pushed a branch with the rebase over here: https://github.com/smlx/saml2aws/tree/fido-linux-support

cc @ocraviotto and @wolfeidau

ocraviotto:
@wolfeidau I've gone ahead and refactored my initial PR to reflect all of your changes to the release flow with goreleaser. Had a couple of initial failures but finally managed to make it work. I've basically split the build/release process between linux and non-linux in Ubuntu and MacOs contexts respectively. This should maintain the original context for non-linux and build and so support u2f keys in linux by building on Ubuntu. I've also changed the Makefile to allow building with goreleaser locally, except that it won't build linux if in non-linux.

I'm not sure whether MacOs binaries would work for all cases if built on Ubuntu (asked a colleague to test and the limited cases he tried did work, but can't be sure without further testing), in which case it would be possible to move the build/release fully to ubuntu-latest (not sure why you had builds on ubuntu and Releases done with macos?). 

Finally, I've changed the Build step from `macOS-latest` to `macos-latest` (aligned case to [the docs specs for OS types supported and their YAML label](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on)), so I suspect that's why the PR check is missing: https://github.com/Versent/saml2aws/pull/592/checks?check_run_id=2295884756.

Do let me know if this now looks more in line with what you wanted and of course I'd be happy to read your feedback.

ocraviotto:
@smlx Tanks for trying it out. Please note I've rebased and updated with changes that should allow you to build with U2F keys support from linux. I've created a test release in my fork with almost the same code base I used to update this PR. Release builds here: https://github.com/ocraviotto/saml2aws/releases/tag/v2.28.5-testrel1

wolfeidau:
@ocraviotto yeah I have dug into this and agree it is 100% already there and actively used.

I need to pull apart your PR a bit as a lot of things are going in there.

ocraviotto:
@wolfeidau Rebased once more. I see actions need a maintainer's approval to run, might be a good additional check. I have fixed a minor conflict but have not noticed anything else worth checking in more detail, though did run tests and builds locally. 
As earlier, if there is anything I can do to help, happy to.

.github/workflows/release.yml
```diff
@@ -8,7 +8,10 @@ on:
 jobs:
   release:
     name: release
-    runs-on: macOS-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
duboisf, line 14: would it work to run only on `ubuntu-latest` and have a single `.goreleaser.yml` that has 2 builds, something like:

```
builds:
- id: saml2aws-linux
  main: ./cmd/saml2aws/main.go
  binary: saml2aws
  flags:
    - -tags=hidraw
    - -trimpath
    - -v
  ldflags:
    - -s -w -X main.Version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
  goos:
    - linux
  goarch:
    - amd64
    - arm64
    - arm
- id: saml2aws-non-linux
  main: ./cmd/saml2aws/main.go
  binary: saml2aws
  flags:
    - -trimpath
    - -v
  ldflags:
    - -s -w -X main.Version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
  goos:
    - windows
    - darwin
  goarch:
    - amd64
    - arm64
    - arm
```

You would also avoid having to conditionally install libudev-dev and also avoid having to use a matrix.

duboisf: hmmmm scratch that, you can't easily cross-compile with CGO enabled :disappointed:

.github/workflows/go.yml
```diff
@@ -50,7 +59,10 @@ jobs:
 
   release-build:
     name: release-build
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
duboisf, line 65: you could do something like

```
strategy:
  matrix:
    include:
    - os: macos-latest
      build-type: non-linux
    - os: ubuntu-latest
      build-type: linux
```

then you can use it like:

```
    - name: GoReleaser
      uses: goreleaser/goreleaser-action@v2
      with:
        version: latest
        args: build --snapshot --rm-dist --config .goreleaser-${{ matrix.build-type }}.yml
```

.github/workflows/go.yml
```diff
@@ -61,8 +73,20 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: GoReleaser
+    - name: Install libudev if in Linux
+      run: sudo apt-get install libudev-dev
+      if: ${{ startsWith(matrix.os, 'ubuntu') }}
```
duboisf, line 78: you can move this into `.goreleaser-linux.yml`:

```
before:
  hooks:
    - sudo apt-get install libudev-dev
```

jplanckeel:
Hello, we have test fork release for FIDO U2F on linux Fedora, it works perfectly.

When you can integrated this in release ?

abualy:
worked perfectly for fedora :D
libudev-dev quivalent in fedora is ... systemd-devel
```shell
dnf install systemd-devel
```

rarruda:
Anything I can do to help to get this merged?

I would love to be able to use it in linux (and i could use a custom build myself, as hack), but I don't want to recommend it to others in my team unless it merged upstream.

vascorsd:
Hello, as an interested party, any reason why this is still pending and support for webauthn not merged yet? From the comments seems the PR works. What are the pending concerns?

PR #593, opened 2021-01-15
Move login details validation to provider

adammw:
Finally got sick of having to put a dummy password in to saml2aws when using the shell provider, so moving the validate into the provider so they can be different. Not sure how it interacts with iDP validate method or how they're different...


adammw:
cc @wolfeidau

wolfeidau:
Going to test this out a bit, I LOVE the idea of making it more pluggable though, there are a few places where complexity has grown with the number of providers.

PR #595, opened 2021-01-20
Adding Credential Process support during login

sledigabel:
`Credential process` is a convenient way to interface a credential provider with the AWS CLI.
https://github.com/awslabs/awsprocesscreds

This change adds support of the credential process format through an extra flag during login.
It also adds a `--quiet` flag to silence the logging output and keep only the output from the credential process.


danmx:
cc @wolfeidau

wolfeidau:
Thanks will be handy for people.

tailrecur:
@sledigabel Thanks a lot for this !!
@wolfeidau Would it be possible to cut a release including this feature?

PR #602, opened 2021-02-08
Adding option to write to different cred file.

sledigabel:
The AWS SDK reads to the default AWS credential file, but does not
handle natively the expiry timestamp. The credential_process allows the
cred renewals when needed, but as the credential file takes precedence,
the credential provider never triggers.

This feature allows saml2aws to write to a different credential file,
which it will use as credential cache and not "conflict" with the AWS
SDK.

The change includes the command line option and the saml2aws
configuration file item.


danmx:
cc @wolfeidau

pkg/awsconfig/awsconfig.go
```diff
@@ -42,9 +42,10 @@ type CredentialsProvider struct {
 }
 
 // NewSharedCredentials helper to create the credentials provider
-func NewSharedCredentials(profile string) *CredentialsProvider {
```
oshaughnessy, line 45: @sledigabel, I believe saml2aws supports `AWS_SHARED_CREDENTIALS_FILE` pretty well. I've been setting that before invoking `saml2aws`, but having this built into the app would be pretty handy. Perhaps it would make sense to support it in `locateFileConfig()` (instead of here? not sure) so that it interacts well with the way that env var is used?

sledigabel: You're right, it does support it and `saml2aws` already supports it in the said function :-)
https://github.com/Versent/saml2aws/blob/master/pkg/awsconfig/awsconfig.go#L174

As mentioned in the PR, the AWS SDK will also use that same env variable, and thus `saml2aws` and any aws sdk tool like `awscli` would point to the same file. It would not be a problem if the sdk would handle expired credentials, but they do not. The credential process (in PR #595) fixes this problem, however the credentials take precedence over the credential process.

I could work around the issue by having a wrapper that sets `AWS_SHARED_CREDENTIALS_FILE` for `saml2aws` but that makes it harder to distribute vs an additional flag set once in the `configure` process.

Another use case is the use of multiple `IDP account`, they would conflict between themselves (writing into the same file) and the feature would allow to distinguish credentials from different accounts.

sledigabel: Re-read your comment @oshaughnessy, I think I understand what you mean now.
`locateFileConfig()` will still be instanciated, and will honour the `AWS_SHARED_CREDENTIALS_FILE` as before if a specific credentials file is not set up (the default mode). That allows the ini file, and the CLI option and the specific SAML2AWS env variable to superseed the `AWS_SHARED_CREDENTIALS_FILE` if they are present.

Thank you for pointing this out, I'll add this to the documentation.

oshaughnessy: Yeah, thanks for clarifying. I have been using a wrapper to set `AWS_SHARED_CREDENTIALS_FILES`, but I don't like it for the same reasons you mentioned. This would get me one step closer to only needing to distribute a configuration script.

sledigabel:
@wolfeidau what do you think? are you happy with the way the credential-file is parameterised?

wolfeidau:
This is released as `v2.28.2` keen to hear how it goes.

Cheers.

oshaughnessy:
Thanks! I'll test this out, too.

PR #607, opened 2021-02-13
Adding SAML Assertion caching feature

### LINKED ISSUE #526: New feature: Briefly caching SAML Assertions
I'm interested in caching of the full SAML assertion that is received from the IDP. I could contribute a PR for this, just wanted to open a ticket first to gather agreement on approach, storage location, enablement strategy, etc.

I'm using the OneLogin provider, and their API does not allow for reusing a MFA'd session, so each SAML flow requires a fresh MFA code. However the OneLogin SAML assertions themselves are valid for **6 minutes**.

During this window the SAML assertion can be sent to AWS repeatedly to create multiple sessions to different AWS accounts and AWS roles. After it expires then it needs to just be thrown away and the user prompted to perform a fresh login flow.

This cache wouldn't be as long-lived as a proper IDP session cache (as requested in #492) however it should work the same way with every SAML provider so it should be immediately usable by anyone, not just users of a certain provider.

As far as implementation, I think placing the assertion on disk (as opposed to keychain) makes sense due to its larger size and short livespan. OneLogin's own CLI tool [added SAML caching](https://github.com/onelogin/onelogin-python-aws-assume-role/commit/9e8137b7606d315ea1c491c093e689e38c5a16b3) as an opt-in feature and stores the body at `$HOME/.aws/saml_cache.txt` (saml2aws would likely want to use a dynamic filename that includes the provider config name, if filesystem is used)

sledigabel:
You can now cache the SAML assertion after authentication using the flag
`--cache-saml`. You can use the flag at `configure` time and it will be
set on subsequent `login` and `list-roles` calls.

The timeout for a SAML assertion is set to 5min as a constant, and the
code makes it relatively easy to make it configurable if needed, like
for example override this per provider.

Fixes #526

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,7 +635,9 @@ When using the aws cli with the `mybucket` profile, the authentication process w
 
 You can use the flag `--cache-saml` in order to cache the SAML assertion at authentication time. The SAML assertion cache has a very short validity (5 min) and can be used to authenticate to several roles with a single MFA validation.
 
-The cache name file is called `.saml2aws_cache` and is stored in your `.aws` directory in your user homedir.
+there is a file per saml2aws profile, the cache directory is called `.saml2aws_cache` and is located in your `.aws` directory in your user homedir.
+
+You can toggle `--cache-saml` during `login` or during `list-roles`, and you can set it once during `configure` and use it implicitly.
 
 # License
 
--- a/README.md
+++ b/README.md
@@ -635,7 +635,7 @@ When using the aws cli with the `mybucket` profile, the authentication process w
 
 You can use the flag `--cache-saml` in order to cache the SAML assertion at authentication time. The SAML assertion cache has a very short validity (5 min) and can be used to authenticate to several roles with a single MFA validation.
 
-there is a file per saml2aws profile, the cache directory is called `.saml2aws_cache` and is located in your `.aws` directory in your user homedir.
+there is a file per saml2aws profile, the cache directory is called `saml2aws` and is located in your `.aws` directory in your user homedir.
 
 You can toggle `--cache-saml` during `login` or during `list-roles`, and you can set it once during `configure` and use it implicitly.
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ The process goes something like this:
 * Prompt user for credentials
 * Log in to Identity Provider using form based authentication
 * Build a SAML assertion containing AWS roles
+* Optionally cache the SAML assertion
 * Exchange the role and SAML assertion with [AWS STS service](https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html) to get a temporary set of credentials
 * Save these credentials to an aws profile named "saml"
 
@@ -629,6 +630,12 @@ credential_process = saml2aws login --skip-prompt --quiet --credential-process -
 ```
 
 When using the aws cli with the `mybucket` profile, the authentication process will be run and the aws will then be executed based on the returned credentials.
+
+# Caching the saml2aws SAML assertion for immediate reuse
+
+You can use the flag `--cache-saml` in order to cache the SAML assertion at authentication time. The SAML assertion cache has a very short validity (5 min) and can be used to authenticate to several roles with a single MFA validation.
+
+The cache name file is called `.saml2aws_cache` and is stored in your `.aws` directory in your user homedir.
 
 # License
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The process goes something like this:
 * Prompt user for credentials
 * Log in to Identity Provider using form based authentication
 * Build a SAML assertion containing AWS roles
-* Optionally cache the SAML assertion
+* Optionally cache the SAML assertion (the cache is not encrypted)
 * Exchange the role and SAML assertion with [AWS STS service](https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html) to get a temporary set of credentials
 * Save these credentials to an aws profile named "saml"
 
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -32,6 +32,8 @@ func Login(loginFlags *flags.LoginExecFlags) error {
 	}
 
 	sharedCreds := awsconfig.NewSharedCredentials(account.Profile, account.CredentialsFile)
+	// creates a cacheProvider, only used when --cache is set
+	cacheProvider := samlcache.NewSAMLCacheProvider(account.Profile, "")
 
 	logger.Debug("check if Creds Exist")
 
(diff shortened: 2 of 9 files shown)


sledigabel:
The test is failing because the cache file cannot be stored in the default path in Github actions.
Will refactor the code for better testability.

sledigabel:
Just found a bug, the account.Profile is in fact the aws profile not the saml one. need to push one more fix.

sledigabel:
This is ready for review @wolfeidau :-)

pkg/samlcache/samlcache.go
```diff
@@ -0,0 +1,155 @@
...
+var (
+	ErrInvalidCachePath = errors.New("Cannot evaluate Cache file path")
+	logger              = logrus.WithField("pkg", "samlcache")
+)
+
+const (
+	SAMLAssertionValidityTimeout = 5 * time.Minute
+	SAMLCacheFilePermissions     = 0600
+	SAMLCacheDirPermissions      = 0700
+	SAMLCacheDir                 = ".saml2aws_cache"
```
danmx, line 26: AWS CLI is using [`.aws/cli/cache`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) for its cache directory. If we want to use `.aws` directory then using something like `.aws/saml2aws/cache` would make more sense.

sledigabel: no problem, that can be changed.
On the customization front, the code allows it but it's already a pretty dense PR and would prefer address this in a separate PR, if there's a need for it.

README.md
```diff
@@ -12,6 +12,7 @@ The process goes something like this:
 * Prompt user for credentials
 * Log in to Identity Provider using form based authentication
 * Build a SAML assertion containing AWS roles
+* Optionally cache the SAML assertion
 * Exchange the role and SAML assertion with [AWS STS service](https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html) to get a temporary set of credentials
```
wolfeidau, line 16: I probably call out this is not encrypted.

cmd/saml2aws/commands/login.go
```diff
@@ -31,6 +32,8 @@ func Login(loginFlags *flags.LoginExecFlags) error {
 	}
 
 	sharedCreds := awsconfig.NewSharedCredentials(account.Profile, account.CredentialsFile)
+	// creates a cacheProvider, only used when --cache is set
+	cacheProvider := samlcache.NewSAMLCacheProvider(account.Name, "")
```
wolfeidau, line 37: Is the second argument to `NewSAMLCacheProvider` always an empty string?

cmd/saml2aws/commands/login.go
```diff
@@ -79,10 +82,30 @@ func Login(loginFlags *flags.LoginExecFlags) error {
...
 
-	samlAssertion, err := provider.Authenticate(loginDetails)
-	if err != nil {
-		return errors.Wrap(err, "error authenticating to IdP")
+	var samlAssertion string
+	if account.SAMLCache {
+		if valid, err := cacheProvider.IsValid(); valid {
+			samlAssertion, err = cacheProvider.Read()
+			if err != nil {
+				logger.Debug("Could not read cache:", err)
```
wolfeidau, line 90: There probably should be a `ErrNotFound` rather than just skipping all errors and continuing, kinda makes the error redundant if you always ignore it.

cmd/saml2aws/commands/login.go
```diff
@@ -79,10 +82,30 @@ func Login(loginFlags *flags.LoginExecFlags) error {
 
 	log.Printf("Authenticating as %s ...", loginDetails.Username)
 
-	samlAssertion, err := provider.Authenticate(loginDetails)
-	if err != nil {
-		return errors.Wrap(err, "error authenticating to IdP")
+	var samlAssertion string
+	if account.SAMLCache {
+		if valid, err := cacheProvider.IsValid(); valid {
```
wolfeidau, line 87: Your ignoring the error here, if there is an error you should probably log it and stop

cmd/saml2aws/commands/login.go
```diff
@@ -79,10 +82,30 @@ func Login(loginFlags *flags.LoginExecFlags) error {
...
+	if samlAssertion == "" {
+		// samlAssertion was not cached
+		samlAssertion, err = provider.Authenticate(loginDetails)
+		if err != nil {
+			return errors.Wrap(err, "error authenticating to IdP")
+		}
+		if account.SAMLCache {
+			err = cacheProvider.Write(samlAssertion)
+			if err != nil {
+				logger.Error("Could not write samlAssertion:", err)
```
wolfeidau, line 106: Missing a return?

wolfeidau (changes requested):
Just some general code feedback.

cmd/saml2aws/commands/login.go
sledigabel: that's right;
It's mimicking `NewSharedCredentials` and is a placeholder that will allow parameterising the file location in the future. I thought it would be too heavy to bundle this in this PR and not sure how the cli would look like either.
At the moment it's empty string, which means we'll always use the default location.

I'm happy to remove the extra parameter if it adds confusion.

README.md
sledigabel: good shout

cmd/saml2aws/commands/login.go
sledigabel: good call. will fix

cmd/saml2aws/commands/login.go
sledigabel: good call, i'll actually error on read failure

cmd/saml2aws/commands/login.go
sledigabel: I had designed it in a non-blocking way, but I get your point. If there's an intention to use cache and we can't write it, we should error.
will add 👍

cmd/saml2aws/commands/login.go
sledigabel: Considered what you said and simplified the SAMLCacheProvider creation altogether. We can add the filename later if we want to as easily.

cmd/saml2aws/commands/login.go
sledigabel: refactored `IsValid` to returning a bool only

sledigabel:
@wolfeidau I've addressed your feedbacks, thanks a lot for these.
Let me know if they look better now

wolfeidau (approved):

wolfeidau:
@sledigabel Thanks for all the work and patience, would be great if you could test out the latest release when you get a moment.

sledigabel:
@wolfeidau I've tested this successfully using ADFS.
Thanks a lot for merging and releasing!
Quick question how does the new version arrive in brew?

PR #616, opened 2021-02-16
Change macOS install instruction to install from homebrew-core

crunchtime-ali:
Congratulations! `saml2aws` has been added to [homebrew-core](https://github.com/Homebrew/homebrew-core). Therefore, it can be installed right away with `brew install saml2aws` without having to add a tap.

Do you also want me to add a `tap_migrations.json` file to your own tap that will automatically migrate all installations to `homebrew-core`? (see https://docs.brew.sh/Migrating-A-Formula-To-A-Tap)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ Aside from Okta, most of the providers in this project are using screen scraping
 If you're on OSX you can install saml2aws using homebrew!
 
 ```
-brew tap versent/homebrew-taps
 brew install saml2aws
 saml2aws --version
 ```


wolfeidau:
@crunchtime-ali The current tap is here https://github.com/Versent/homebrew-taps/blob/master/saml2aws.rb would love some help migrating. Any tips or examples appreciated.

crunchtime-ali:
@wolfeidau It was very simple. See here https://github.com/Versent/homebrew-taps/pull/27

Since `goreleaser` is not configured to update your own homebrew tap I think there is nothing else to do at this time.
Thank you for developing this tool. It is among the most important ones for my team 🙇

PR #618, opened 2021-02-17
readme: Added profile in credential process

lamp0chka:
It seems that by skipping --profile flag in credential_process authentication, saml2aws will use [saml] profile in `~/.aws/credentials` to store the temp keys. 

This could lead to the situation where person was authenticated in account A, credentials hasn't expired yet -> they are not renewed -> account A details are used, even though we want to access account B.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,7 +616,7 @@ An example of the aws configuration (`~/.aws/config`):
 ```
 [profile mybucket]
 region = us-west-1
-credential_process = saml2aws login --skip-prompt --quiet --credential-process --role <ROLE>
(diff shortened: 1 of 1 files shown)


sledigabel:
Good catch @lamp0chka thank you!

sledigabel:
If you've been using the `--profile` as I did, have a look at my other PR (#602)
You'll find that credential_process doesn't work once the credentials have been produced and have expired.
Once merged, you'll be able to have `saml2aws` write to another credential file and have the credential-process completely transient.

PR #621, opened 2021-02-19
fix(build): do some work on darwin build flags to enable build anywhere

wolfeidau:


oshaughnessy:
@wolfeidau, I've been trying local builds on CentOS 7 and I get a failure when trying to add in bearsh/hid:

```
# github.com/bearsh/hid
In file included from ../../../../pkg/mod/github.com/bearsh/hid@v1.3.0/hid_enabled.go:30:0:
../../../../pkg/mod/github.com/bearsh/hid@v1.3.0/libusb/libusb/os/threads_posix.c: In function 'usbi_get_tid':
../../../../pkg/mod/github.com/bearsh/hid@v1.3.0/libusb/libusb/os/threads_posix.c:86:23: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'unsigned'
 static _Thread_local unsigned int tl_tid;
<winmail.dat>
```

I have gcc 4.8.5. Any idea what might be causing that?

wolfeidau:
@oshaughnessy yeah one of the deps used for access to ubikey / USB Second factors is using this HID library. 

I build it fine on ubuntu 20.04 so I am guessing that libusb version is the issue, new methods ect which are missing in your version on Centos7

We would need to wrap that stuff in a build flag to disable it for your use case i think.

Issue #507, opened 2020-05-27
Okta Verify MFA prompts twice


hiruna:
Here is a copy of my config
```
account {
  URL: <okta_aws_app_url>
  Username: <my_email>
  Provider: Okta
  MFA: Auto
  SkipVerify: false
  AmazonWebservicesURN: urn:amazon:webservices
  SessionDuration: 3600
  Profile: default
  RoleARN: 
  Region: 
}
``` 
Upon executing `saml2aws login`, the cli prompts me for the MFA twice before listing the roles:
![image](https://user-images.githubusercontent.com/3973928/82966260-d3674180-a00d-11ea-9fe4-bd053eab599b.png)

This also occurs when using the code verification option (instead of the push).

steverukuts:
I encountered this issue, when Okta was configured to require MFA again before loading an AWS role. Even if you set the timeout in the UI to give you 10 minutes between MFA accesses, it would do that. To work around, you can disable that feature.

To fix it in the code, perhaps the MFA response could be repeated twice in this case and it might work.

jayzalowitz:
@steverukuts How do you de-configure okta to require the mfa before loading?

steverukuts:
It'll depend on your setup. To find out the config, navigate to your app, go to signon and look at your sign-on policy rules. Your organisation might have made some accounts re-require MFA. You can add a new rule that doesn't require re-prompting for MFA.

Note that there can be a lot of variance between organisation accounts and what you bought etc. This might also be adaptive MFA being suspicious about your activity. If you can't figure it out, the Okta support team are pretty helpful in situations like this.

jayzalowitz:
Can you hit me up at jayzalowitz at googles normal non german mail domain.com so I can confirm something? Im pretty sure our rules are following that and its still happening.

jayzalowitz:
I got pretty blanket allow access set up here and its not playing nice.

steverukuts:
I'd love to help but I'm afraid I'm about to sign off for the weekend. Okta's support team can certainly solve this problem for you. It costs enough so you should make them work for their money.

jayzalowitz:
Hah, to be honest they aren't even showing up, 1/10 would not recommend.

matrwall:
We have a similar issue for staff accessing the AWS account/roles outside of the network.
We have it configured for PUSH MFA by default in our config.

Okta is configured for MFA Authentication at 2 levels outside of the office network:
[...]

Issue #492, opened 2020-04-25
Caching IDP (Okta) Sessions


kaitea:
When running `saml2aws login` , it keeps asking for the MFA token for every new login attempt (to a different aws account)

Is there a way to cache the login session to IDP (okta) and not have to enter an MFA token for login attempts to different aws accounts? (Our Okta setup allows for 24hr login session, after password and MFA token entry)

Thanks

scottgerring:
I'd also be interested in this! 
If there's a clean way of implementing it within the tool is it something you'd take a PR for?

nickchappell:
I'd be interested in this as well. `okta-aws` currently does this, via cached tokens in ~/.okta/cookies.properties`

https://github.com/oktadeveloper/okta-aws-cli-assume-role

taylorsmcclure:
This would help out UX immensely. I have an Amazon Web Services "sign on" rule that requires an MFA challenge once per session. It requires an Okta double push which confuses and irritates the end-users.

curator:
My company has it's own, much modified version of https://github.com/RedVentures/oktad that we're trying to get away from, but the caching is a pretty serious need for our workflows.  Most of our teams work in 2-5 accounts a day and regularly for-loop auth to them, but having to do the MFA dance on each every account (sometimes as many as a dozen) is exceptionally tedious.  

Love this tool otherwise and would love to see this feature.

hoangminhtu-dh:
I think this is already implemented in the most recent version of this tool? Today I tried it out and it stopped asking me for Okta OTP in my multiple, sequential saml2aws runs.

PR #627, opened 2021-03-01
Fix "Verify device" double prompt for Okta

alsmola:
I believe this solves the issue reported in https://github.com/Versent/saml2aws/issues/507 by adding a Cookie Jar to the HTTP client and making a dummy request to the Okta App URL get a device token (`dt`) cookie from Okta.

It mirrors the approach from `aws-okta` - see https://github.com/segmentio/aws-okta/blob/master/lib/okta.go#L240

I could try to parse out the `dt` cookie (and any others - I think at least `sid` could be required) and thread them through manually, but it seems to me that keeping a Cookie Jar for this client is the right approach, since it's performing a multi-step interaction emulating a web browser.

This doesn't persist the Cookie Jar between runs like `aws-okta` does, so it doesn't directly address https://github.com/Versent/saml2aws/issues/492

### DIFF
--- a/pkg/provider/okta/okta.go
+++ b/pkg/provider/okta/okta.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/http/cookiejar"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


wolfeidau:
@alsmola nice work, we have a tenancy which exhibits this issue as well will give it a go.

Thanks for figuring it out!

PR #564, opened 2020-09-25
Support multiple fido devices in Okta

dustinblackman:
## Overview

Okta allows you to attach multiple Fido devices to your account, but `saml2aws` will only attempt the use the first device in the list returned from Okta. This results in an authentication failure if the local device does not match, with no way to select another device. This PR iterates through each Fido device on the account looking for a match on the local system. 

This required a slight refactor to make requests to Okta's verification endpoints more reusable, but will make future refactoring easier.

PR #630, opened 2021-03-11
dustinblackman okta multi fido

wolfeidau:
I have taken the code from #564 and refactored / adjusted it a little.

## Overview

Okta allows you to attach multiple Fido devices to your account, but `saml2aws` will only attempt the use the first device in the list returned from Okta. This results in an authentication failure if the local device does not match, with no way to select another device. This PR iterates through each Fido device on the account looking for a match on the local system. 

This required a slight refactor to make requests to Okta's verification endpoints more reusable, but will make future refactoring easier.


wolfeidau:
@dustinblackman I made sure i kept your original commit and just refactored from there, this ensures the main chunk of the change is still attributed to you even though i rebased / refactored.

Issue #600, opened 2021-02-05
Google prompts: unable to find supported second factor


donofriov:
I just updated to the latest version of `saml2aws` and am no longer able to authenticate via Google prompts which was working before the update.

```
Authenticating as vincent.donofrio@soul-cycle.com ...
error authenticating to IdP: error loading challenge page: unable to find supported second factor
```

```
macOS Big Sur Version 11.1 (20C69)
❯ saml2aws --version
2.27.1
❯ aws --version
aws-cli/2.1.24 Python/3.9.1 Darwin/20.2.0 source/x86_64 prompt/off
❯ go version
go version go1.15.7 darwin/amd64
❯ brew --version
Homebrew 3.0.0
Homebrew/homebrew-core (git revision 8fd6ba; last commit 2021-02-05)
Homebrew/homebrew-cask (git revision 56eb4f; last commit 2021-02-05)
```

<img width="717" alt="image" src="https://user-images.githubusercontent.com/4457953/107035496-8d779580-67b8-11eb-91cd-2a5dd3f28554.png">

`❯ saml2aws login -a nonprod-saml-admin --verbose`
```
DEBU[0000] Running                                       command=login
DEBU[0000] check if Creds Exist                          command=login
DEBU[0000] Expand                                        name=/Users/vdonofrio/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/Users/vdonofrio/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/Users/vdonofrio/.aws/credentials pkg=awsconfig
Using IDP Account nonprod-saml-admin to access GoogleApps https://accounts.google.com/o/saml2/initsso?idpid=<REDACTED>&spid=<REDACTED>&forceauthn=false
DEBU[0000] Get credentials                               helper=osxkeychain serverURL="https://accounts.google.com/o/saml2/initsso?idpid=<REDACTED>&spid=<REDACTED>&forceauthn=false"
DEBU[0000] Get credentials                               helper=osxkeychain user=vincent.donofrio@soul-cycle.com
To use saved password just hit enter.
? Username vincent.donofrio@soul-cycle.com
? Password **************

[...]

PR #633, opened 2021-03-16
Issue 600: Fix Google Prompts

donofriov:
Fixes https://github.com/Versent/saml2aws/issues/600

### DIFF
--- a/pkg/provider/googleapps/googleapps.go
+++ b/pkg/provider/googleapps/googleapps.go
@@ -413,7 +413,7 @@ func (kc *Client) loadChallengePage(submitURL string, referer string, authForm u
 			responseForm.Set("TrustDevice", "on")
 
 			return kc.loadResponsePage(secondActionURL, submitURL, responseForm)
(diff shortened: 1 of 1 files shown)


donofriov:
Hey @wolfeidau I'm not sure of the review / release process. Could you take a look at this small change? Let me know if you need any other information. Thanks!

mattgialelis:
Hey @donofriov I had actually tried this myself originally but it hadnt worked due to the code trying to gather fields that don't exist for the challenge/dp 

Along with attempting to wait for a response from this url to hold the login process until a response was sent
https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L428

I actually just raised a PR with a solution 

https://github.com/Versent/saml2aws/pull/637

The comment should explain how it works but basically all the challenge/dp requires is to refresh the page or attempt to load the next one to see if the challenge was completed if not it loops you back to the same page.

There where a few Data fields required from the form for it but they are already captured earlier on in the code and stored in the response form.

PR #634, opened 2021-03-16
Adding SAML cache file path option

sledigabel:
Extra fixing of #526, allows to override the SAML cache file location in lieu of the default one.
Will allow different IDPAccount configurations to share a single SAML cache file.

Added to the documentation.
Fixed a Marshalling inconsistency with IDPAccount where Name is stored with
a capital lettter.


max-rocket-internet (approved):
Nice, thanks!

sledigabel:
tagging @wolfeidau for visibility

PR #641, opened 2021-03-25
Adding ENV as a script output

sledigabel:
`script` is currently structured around shell compatible outputs only.
Some tools handle native env files, while looking similar they are
syntactically different.

Refactored the script code for testability.

Added some documentation.


wolfeidau:
👍 great work, love the test!

sledigabel:
thank you Mark 🙇  any chance you could cut a new release?

PR #659, opened 2021-04-23
build releases with U2F support on linux

JTarasovic:
## Background

U2F device enumeration for JumpCloud is via [marshallbrekka/go-u2fhost](https://github.com/marshallbrekka/go-u2fhost) which requires `-tags=hidraw` to be built with raw HID support otherwise it fails to find the U2F device.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,6 +7,7 @@ builds:
   flags:
     - -trimpath
     - -v
+    - -tags={{ if (eq .Env.GOOS "linux") }}hidraw{{ end }}
   ldflags:
     - -s -w -X main.Version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
   goos:


JTarasovic:
And here's it working after a `goreleaser build`:
```
❯  ~/Development/saml2aws/dist/saml2aws_linux_amd64/saml2aws list-roles
Using IDP Account default to access JumpCloud https://sso.jumpcloud.com/saml2/aws
To use saved password just hit enter.
? Username
? Password *********************************


Touch the flashing U2F device to authenticate...
  ==> Touch accepted. Proceeding with authentication
  ```

JTarasovic:
may be partially or totally obviated by #592

wolfeidau:
This is a super interesting idea, I def need to exclude code from some builds at the moment as they are failing on windows due to some of this code.

JTarasovic:
shoot. I didn't realize the actions ran on osx. Not sure that there will be a reasonable way to resolve as `libudev` is linux specific...

ocraviotto:
@JTarasovic 

> shoot. I didn't realize the actions ran on osx. Not sure that there will be a reasonable way to resolve as `libudev` is linux specific...

This is the reason I separated the builds. Another option, depending on whether the MacOS build on Ubuntu is fully functional (and the same for Windows), would be to only build on Ubuntu in which case your simpler change would work.

JTarasovic:
@ocraviotto yeah, that'd work for the Action but it'd still :hankey: the bed for devs running on OSX that wanted to test future goreleaser changes, right? Not sure how big of a deal that is. :thinking:

PR #660, opened 2021-04-27
Make providers compatible with non-AWS targets

mjg59:
Nothing about the SAML response provided by the identity providers is tied
to Amazon other than the redirect URL and the provider ID. The provider ID
can be overridden with the aws_urn config parameter, but for some providers
the redirect URL is hardcoded in order to detect a page containing the SAML
response. This patch adds a target_url configuration parameter to allow the
hardcoded values to be overridden, making it possible to use the providers
for non-AWS purposes.


wolfeidau:
@mjg59 This is a pretty interesting change, I had never considered saml2aws not being used for AWS 🤯

mjg59:
Whoops, looks like I didn't run the entire test suite. go test completes locally now.

Issue #455, opened 2020-03-05
Browser based login


anthonyangel:
Is there a possibility of the CLI opening up a browser page for a UI login? (the equivalent of what the `OKTA_BROWSER_AUTH ` flag would achieve in https://github.com/oktadeveloper/okta-aws-cli-assume-role

I'd have a preference for logging in with a browser rather than copying my password into a terminal (or saving it)

jarmbruster74:
Went looking for this as well - try saml2aws console
Love it when it's already there :-)

anthonyangel:
@jarmbruster74 That command will launch an authenticated console session in a browser from the CLI - what I was searching for here is different, it would be a way of starting the login from from the CLI, but then providing authentication credentials via a browser (so that I can use a password manager)

clausa:
A browser based login would also allow for IdP brokering setups. 
If you for instance, have KeyCloak configured as an identity broker and redirects authentication requests to G-Suite (or other IdPs), saml2aws will fail.

musabmasood:
We're behind Okta for a lot of stuff, this would definitely help pasting passwords/codes in the terminal.

CoreySmithRing:
@wolfeidau is there any supporting documentation on how to leverage this browser for login?

mvgijssel:
Would love this as well! We are coming from https://github.com/segmentio/aws-okta which opened a browser window for OKTA authentication.

merusso:
The implementation for this feature downloads independent browser binaries to a cache directory, then launches Chromium.

```
2023/01/18 10:01:36 Downloading driver to /Users/mrusso/Library/Caches/ms-playwright-go/1.14.0
2023/01/18 10:01:38 Downloaded driver successfully
2023/01/18 10:01:38 Downloading browsers...
Downloading Playwright build of chromium v907428 - 117.5 Mb [====================] 100% 0.0s 
Playwright build of chromium v907428 downloaded to /Users/mrusso/Library/Caches/ms-playwright/chromium-907428
Downloading Playwright build of ffmpeg v1005 - 1.3 Mb [====================] 100% 0.0s 
Playwright build of ffmpeg v1005 downloaded to /Users/mrusso/Library/Caches/ms-playwright/ffmpeg-1005
Downloading Playwright build of firefox v1281 - 74.7 Mb [====================] 100% 0.0s 
Playwright build of firefox v1281 downloaded to /Users/mrusso/Library/Caches/ms-playwright/firefox-1281
[...]

PR #661, opened 2021-04-30
New browser based login using playright

wolfeidau:
This provides a solution for #455 and enables users to trial and debug issues during the login process.


CoreySmithRing:
I noticed that the cli still prompts for a username/password during a "login" even though its grabbing saml assertion back from the browser. Is this expect behavior?

Issue #622, opened 2021-02-22
error authenticating to IdP: error verifying MFA: The provided key handle is not present on the device, or was created with a different application parameter.


carloshmiranda:
Started to get this error after changing to a new laptop, has this happened to some else?

We use Okta as an IDP and Yubikeys for FIDO.

cdchris12:
This is happening to a few different folks in our company; anyone else seeing this?

sriram-clever:
Ran into this issue in my org when we have users with multiple FIDO keys configured. 

f anyone wants to test this https://github.com/Versent/saml2aws/pull/649/files

cdchris12:
We tested this in our organization, and it appeared to be fixed with https://github.com/Versent/saml2aws/pull/630 . We've not had any further issues since the release of `v2.28.4`.

sriram-clever:
Interesting for us v2.28.4 ended up introducing this issue for users, especially if the first FIDO key registered happened to be the one they normally use. The pull request addresses what looks like a bug in the new code.

ocraviotto:
@sriram-clever Yes, #630 introduced it for me as well (linux client) and precisely I realized when updating an [older PR fixing u2f usb detection in linux ](https://github.com/Versent/saml2aws/pull/592)with that change and testing that I had the same issue given that I use multiple keys and the first one (and only connected) was being escaped. 
I had not seen your PR until just now, but followed the issue and found the same change (only reversed for clarity) fixed it, so I added it along multiple other changes I did to update and make my fix to usb detection acceptable. 
The line with my version of the change is [this one](https://github.com/Versent/saml2aws/pull/592/files#diff-8954db63dc41a4273eb76419d66f6a51a1c27f0ad3036041ab0529fd1c9f470eL320).

jseiser:
We just setup saml2aws and our entire team is getting this issue.  We all have n+1 Yubikeys configured to access Okta.  I did setup a temporary MFA setup with Google Authenticator and it worked, so the problem is down to just the Yubikeys.

The Yubi keys works when we access Okta in the web browser, it pops up asking you to tap it.  Not sure if there is some other configuration required.

```
❯ saml2aws --version
2.33.0
```

```
❯ saml2aws --verbose -a redacted login --force
DEBU[0000] Running                                       command=login
DEBU[0000] Check if creds exist.                         command=login
DEBU[0000] Expand                                        name=/Users/justin/.aws/credentials pkg=awsconfig
[...]

PR #668, opened 2021-05-21
catching other hid device errors and prompt for FIDO auth selections

hinling-sonder:
Try to fix the following problems:
* https://github.com/Versent/saml2aws/issues/622 with setup like this on okta:
Screen Shot 2021-05-19 at 1 33 24 AM

* yubikey usb device throwing some strange errors while waiting for user response. From our experience, they are harmless.  Just catching them and wait for the timeouts to kick in.


shedimon:
Hi @wolfeidau recently we've faced an issue with the problem this PR solves. Okta returns the full list of FIDO MFAs and without the ability to choose the factor (let's say Yubikey) `saml2aws` tries to use the first option in the FIDO MFAs and if it's not a said Yubikey it can use fails by getting the `u2fhost.BadKeyHandleError` which stops the process before trying other options. 

Using the code from this PR resolves this by adding an ability to specifically choose which FIDO device to use for each authorization. 

Is there anything we can do/help with to move forward with this approach?

pkg/provider/okta/okta.go
```diff
@@ -85,6 +85,12 @@ type mfaChallengeContext struct {
 	challengeResponseBody string
 }
 
+// mfaOption store the mfa position in response and mfa description
+type mfaOption struct {
```
shedimon, line 89: @hinling-sonder WDYT about extracting the `profile.authenticatorName` from the json and add it to this struct to have a human-readable MFA option in the prompt? I found it really helpful to have this instead of the full `profile` field.

PR #676, opened 2021-06-07
Use Okta Sessions API

hiruna:
* This is to use Okta Sessions API by default when authenticating
* Okta session cookie is stored in keychain/credentials store
* Use of Okta sessions can be disabled via CLI flags/config entry
* Remembers MFA device by default so that users do not get prompted for MFA every time (as long as the session is valid)
* Defaults to the previous authentication method when sessions are disabled by user

Binaries built & tested on `go version go1.16.4 darwin/amd64`
Please see README for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ Commands:
         --config=CONFIG            Path/filename of saml2aws config file (env: SAML2AWS_CONFIGFILE)
         --cache-saml               Caches the SAML response (env: SAML2AWS_CACHE_SAML)
         --cache-file=CACHE-FILE    The location of the SAML cache file (env: SAML2AWS_SAML_CACHE_FILE)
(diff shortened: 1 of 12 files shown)


julienbonastre:
Also tested on Windows10 Pro x64 Build 1803 and Ubuntu/LXMint v19 - 🎉 👍🏻

wolfeidau:
Thanks, this looks really good, just need to road test it!

pwmcintyre:
@hiruna - i patch my saml2aws and this "just works"  ... well done! so cool! 👌

julienbonastre:
> @hiruna - i patch my saml2aws and this "just works" ... well done! so cool! 👌

Yeh @hiruna  is a weapon 😎  🚀  🦦

Issue #685, opened 2021-07-01
Bug: using --credential-process with okta push verify not working


pseudo-su:
I'm trying to use `--credential-process` with okta provider. The `saml2aws` command is outputting extra data/logging info to stdout that is causing issues (even though `--quiet` and `--skip-prompt` are passed as arguments).

AWS CLI Config file (`~/.aws/config`):

```ini
[profile my-acc-Admin]
region = ap-southeast-2
credential_process = saml2aws login --skip-prompt --quiet --credential-process --role=arn:aws:iam::XXX:role/Admin --profile=my-acc-Admin
```

Saml2AWS config file (`~/.saml2aws`):

```
[default]
url         = https://XXX.okta.com/home/amazon_aws/XXX/XXX
username    = username
provider    = Okta
mfa         = PUSH
skip_verify = false
timeout     = 0
aws_urn     = urn:amazon:webservices
```

When the AWS CLI runs the credential process I get the error:

```
Expecting value: line 1 column 1 (char 0)
```

I believe the problem is that the saml2aws CLI is sending more than just the required JSON payload to stdout. EG If I run the saml2aws command directly it's sending this to stdout:

```raw

Waiting for approval, please check your Okta Verify app ...... Approved

{<the credentials>}

```

## Possible solution

1. Remove uses of `fmt.Printf` and instead (presumably?) use the logger https://github.com/Versent/saml2aws/blob/a882ec419d0e47210fa49f5d30c793cd64dc3628/pkg/provider/okta/okta.go#L786
2. Add an extra CLI flag similar to `--quiet` called something like `--log-stderr` that defaults to sending all log output to stderr. when using `credential_process` that would mean the user could still see the output in the console but it wouldn't get sent to stdout and cause issues https://github.com/Versent/saml2aws/commit/3d5f4ffa6c844684a9235ed8e53b8f03a03e959b#diff-700f078cb6e14d20af7c09e73a1507604a85cfb66d6c54db727f45dc21a13d2aR155-R159

EDIT: After looking at the code it seems like it might default to using `os.Stderr` so the `--log-stderr` suggestion might not be helpful.

iciclespider:
Use `--quiet`.

PR #686, opened 2021-07-01
fix: remove fmt.Print* logging

pseudo-su:
https://github.com/Versent/saml2aws/issues/685

### DIFF
--- a/pkg/provider/akamai/akamai.go
+++ b/pkg/provider/akamai/akamai.go
@@ -168,8 +168,7 @@ func (oc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	authStatus := gjson.Get(resp, "status").String()
 	if authStatus != "200" {
 		authFailReason := gjson.Get(resp, "msg").String()
-		fmt.Printf("Login Failed %s\n", authFailReason)
-		logger.Debug("Login Failed:", authFailReason)
+		log.Printf("Login Failed %s", authFailReason)
 		return samlAssertion, errors.Wrap(err, "Login Failure")
 	}
 
@@ -207,8 +206,7 @@ func (oc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 			return samlAssertion, errors.Wrap(err, "error verifying MFA")
 		}
 	} else if mfaStatus == "register" {
-		fmt.Printf("MFA is enabled but not registered for user. Register MFA by accessing EAA IDP from Browser\n")
-		logger.Debug("MFA is enabled but not registered for user")
+		log.Println("MFA is enabled but not registered for user. Register MFA by accessing EAA IDP from Browser")
 		return samlAssertion, errors.Wrap(err, "register mfa by logging to IDP")
 	}
 
--- a/pkg/provider/googleapps/googleapps.go
+++ b/pkg/provider/googleapps/googleapps.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
@@ -175,7 +176,7 @@ func (kc *Client) tryDisplayCaptcha(captchaPictureURL string) (string, error) {
 }
 
 func (kc *Client) iTermCaptchaPrompt(captchaPictureURL string) (string, error) {
-	fmt.Printf("Detected iTerm, displaying URL: %s\n", captchaPictureURL)
+	log.Printf("Detected iTerm, displaying URL: %s\n", captchaPictureURL)
 	imgResp, err := kc.client.Get(captchaPictureURL)
 	if err != nil {
 		return "", errors.Wrap(err, "unable to fetch captcha image")
@@ -186,7 +187,7 @@ func (kc *Client) iTermCaptchaPrompt(captchaPictureURL string) (string, error) {
 	_ = b64Encoder.Close()
 
 	if os.Getenv("TERM") == "screen" {
-		fmt.Println("Detected tmux, using specific workaround...")
+		log.Println("Detected tmux, using specific workaround...")
 		fmt.Printf("\033Ptmux;\033\033]1337;File=width=40;preserveAspectRatio=1;inline=1;:%s\a\033\\\n", buf.String())
 	} else {
 		fmt.Printf("\033]1337;File=width=40;preserveAspectRatio=1;inline=1;:%s\a\n", buf.String())
@@ -195,7 +196,7 @@ func (kc *Client) iTermCaptchaPrompt(captchaPictureURL string) (string, error) {
 }
 
 func simpleCaptchaPrompt(captchaPictureURL string) string {
-	fmt.Println("Open this link in a browser:\n", captchaPictureURL)
+	log.Println("Open this link in a browser:\n", captchaPictureURL)
 	return prompter.String("Captcha", "")
 }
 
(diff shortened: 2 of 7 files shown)


pseudo-su:
This does fix the issue that I was running into, although it does look like `log.Printf` is inserting newlines where there weren't any when using `fmt.Printf`

eg with `fmt.Printf`

```raw
Waiting for approval, please check your Okta Verify app ...... Approved
```

now using with `log.Printf`:
```
Waiting for approval, please check your Okta Verify app ...
.
.
.
 Approved
```

sledigabel (approved):
FWIW, it looks good to me.
This was an oversight from the credential process PR, didn't check that providers would have their own logging as well.
Nice one

pkg/provider/akamai/akamai.go
```diff
@@ -168,7 +168,7 @@ func (oc *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	authStatus := gjson.Get(resp, "status").String()
 	if authStatus != "200" {
 		authFailReason := gjson.Get(resp, "msg").String()
-		fmt.Printf("Login Failed %s\n", authFailReason)
+		log.Printf("Login Failed %s\n", authFailReason)
```
sledigabel, line 171: Actually that one is a duplicate from the line below. you can remove it altogether or delete the next line.

pkg/provider/googleapps/googleapps.go
```diff
@@ -186,16 +187,16 @@ func (kc *Client) iTermCaptchaPrompt(captchaPictureURL string) (string, error) {
 	_ = b64Encoder.Close()
 
 	if os.Getenv("TERM") == "screen" {
-		fmt.Println("Detected tmux, using specific workaround...")
-		fmt.Printf("\033Ptmux;\033\033]1337;File=width=40;preserveAspectRatio=1;inline=1;:%s\a\033\\\n", buf.String())
+		log.Println("Detected tmux, using specific workaround...")
+		log.Printf("\033Ptmux;\033\033]1337;File=width=40;preserveAspectRatio=1;inline=1;:%s\a\033\\\n", buf.String())
```
sledigabel, line 191: nitpicking but This could prob be a `Debug` rather than a `Printf`

pkg/provider/jumpcloud/jumpcloud_webauthn.go
```diff
@@ -140,14 +141,14 @@ func (d *FidoClient) ChallengeU2F() (*JumpCloudResponse, error) {
 					},
 					Token: d.token,
 				}
-				fmt.Printf("  ==> Touch accepted. Proceeding with authentication\n")
+				log.Printf("  ==> Touch accepted. Proceeding with authentication\n")
```
sledigabel, line 144: This could be a `Println`

pkg/provider/jumpcloud/jumpcloud_webauthn.go
```diff
@@ -140,14 +141,14 @@ func (d *FidoClient) ChallengeU2F() (*JumpCloudResponse, error) {
...
 				return responsePayload, nil
 			}
 
 			switch err.(type) {
 			case *u2fhost.TestOfUserPresenceRequiredError:
 				if !prompted {
-					fmt.Printf("\nTouch the flashing U2F device to authenticate...\n")
+					log.Printf("\nTouch the flashing U2F device to authenticate...\n")
```
sledigabel, line 151: Same, Println

pkg/provider/okta/okta.go
```diff
@@ -783,14 +783,14 @@ func verifyMfa(oc *Client, oktaOrgHost string, loginDetails *creds.LoginDetails,
 
 	case IdentifierPushMfa:
 
-		fmt.Printf("\nWaiting for approval, please check your Okta Verify app ...")
+		log.Printf("\nWaiting for approval, please check your Okta Verify app ...")
```
sledigabel, line 786: Same Println

pkg/provider/okta/okta.go
```diff
@@ -783,14 +783,14 @@ func verifyMfa(oc *Client, oktaOrgHost string, loginDetails *creds.LoginDetails,
...
 
 		// loop until success, error, or timeout
 		body := challengeContext.challengeResponseBody
 		for {
 			// on 'success' status
 			if gjson.Get(body, "status").String() == "SUCCESS" {
-				fmt.Printf(" Approved\n\n")
+				log.Printf(" Approved\n\n")
```
sledigabel, line 793: same, Println

pkg/provider/okta/okta.go
```diff
@@ -800,7 +800,7 @@ func verifyMfa(oc *Client, oktaOrgHost string, loginDetails *creds.LoginDetails,
 
 			case "WAITING":
 				time.Sleep(3 * time.Second)
-				fmt.Printf(".")
+				log.Printf(".")
```
sledigabel, line 803: Not sure this is all that useful :-)

pkg/provider/okta/okta.go
```diff
@@ -809,15 +809,15 @@ func verifyMfa(oc *Client, oktaOrgHost string, loginDetails *creds.LoginDetails,
...
 				return "", errors.New("User did not accept MFA in time")
 
 			case "REJECTED":
-				fmt.Printf(" Rejected\n")
+				log.Printf(" Rejected\n")
 				return "", errors.New("MFA rejected by user")
 
 			default:
-				fmt.Printf(" Error\n")
+				log.Printf(" Error\n")
```
sledigabel, line 820: Perhaps this could be a `log.Error`

pkg/provider/okta/okta_webauthn.go
```diff
@@ -111,14 +112,14 @@ func (d *FidoClient) ChallengeU2F() (*SignedAssertion, error) {
...
-				fmt.Printf("  ==> Touch accepted. Proceeding with authentication\n")
+				log.Printf("  ==> Touch accepted. Proceeding with authentication\n")
 				return responsePayload, nil
 			}
 
 			switch err.(type) {
 			case *u2fhost.TestOfUserPresenceRequiredError:
 				if !prompted {
-					fmt.Printf("\nTouch the flashing U2F device to authenticate...\n")
+					log.Printf("\nTouch the flashing U2F device to authenticate...\n")
```
sledigabel, line 122:
```suggestion
					log.Println("\nTouch the flashing U2F device to authenticate...")
```

pkg/provider/onelogin/onelogin.go
```diff
@@ -360,7 +360,7 @@ func verifyMFA(oc *Client, oauthToken, appID, resp string) (string, error) {
 			switch gjson.Get(string(body), "status.type").String() {
 			case TypePending:
 				time.Sleep(time.Second)
-				fmt.Print(".")
+				log.Print(".")
```
sledigabel, line 363: Not that useful to log?

sledigabel (changes requested):
On second look, there's a few adjustments that could be done due to it going to logging now, not stdout :-) but overall this should fix the problem for sure 👍

pkg/provider/okta/okta.go
pseudo-su: I'll happily remove it

pkg/provider/googleapps/googleapps.go
pseudo-su: I'm actually not sure what the impact of changing the logging of the terminal emulator command would be so i've switch those back in order for this PR to be more targetted towards removing log entries. The newline behavior of `log.Printf` would likely make a mess of things.

It's possible that these would need to use `fmt.FPrintf` and explicitly specify the output stream.

pkg/provider/onelogin/onelogin.go
pseudo-su: I've removed this and added a `logger.Debug` so now both `okta.go` and `onelogin.go` don't print out the individual dots but if you turn on debug you will see the output while waiting.

pkg/provider/googleapps/googleapps.go
sledigabel: that's a great point; definitely agree on logging this being not only not useful and probably breaking things. we'd have to see the impact on leaving that one on while using credentials process

pkg/provider/okta/okta.go
sledigabel: Another great point; we use logrus in this file:
https://github.com/Versent/saml2aws/blob/302865c7e362604783069cf124a8f98aefccc72e/pkg/provider/okta/okta.go#L45

I would assume using `logger` everywhere would make more sense, rather than using the log one and losing the context that logrus provides.

PR #720, opened 2021-08-25
Add JumpCloud Protect (PUSH) MFA support

logingood:
JumpCloud have released a new app that they call [JumpCloud
Protect](https://jumpcloud.com/press/introducing-jumpcloud-protect-free-mobile-multi-factor-authentication).

It is a free alternative to DUO for JumpCoud users that provides "push"
second factor. It would be useful to add this in if possible (we are
planning to use it).

### DIFF
--- a/pkg/provider/jumpcloud/jumpcloud.go
+++ b/pkg/provider/jumpcloud/jumpcloud.go
@@ -23,22 +23,25 @@ import (
 )
 
 const (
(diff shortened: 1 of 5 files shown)


logingood:
sorry for bugging @wolfeidau but would you be able to review this pr some time soon ? a few people are relying on this authenticator. I personally use duo but a few colleagues want to use JumpCloud Protect/Push.
Thanks!

PR #723, opened 2021-09-06
Find tokens by regex

peterkowalski:
Some Identity providers add some random stuff to the SAML attribute before/after Principal and Role ARNs. This PR improves the way ARNs are identified and removes this additional stuff.

### DIFF
--- a/aws_role.go
+++ b/aws_role.go
@@ -2,6 +2,7 @@ package saml2aws
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


wolfeidau:
@peterkowalski I would love to see some examples of this "random stuff" so i can understand how this regex works.

Would be good to have a test with some examples.

mapkon (commented):
@peterkowalski Can you add some tests on this PR?

PR #737, opened 2021-10-19
Execute command directly without sh -c

### LINKED ISSUE #236: saml2aws exec doesn't escape arguments passed to /bin/sh
## expected output
```
$ touch generated
$ saml2aws exec --profile staging -- echo "generate[a-z]"
Running command as: [REDACTED]
generate[a-z]
```

## actual output
```
$ touch generated
$ saml2aws exec --profile staging -- echo "generate[a-z]"
Running command as: [REDACTED]
generated
```

rwstauner:
When the args are concatenated there is a loss of fidelity:
arguments with spaces will be split and any special characters
will be evaluated by the shell.
If we simply execute the provided command it will propagate correctly
and it can always be wrapped in a shell on the command line
if that behavior is desired.

fixes #236


wolfeidau (approved):
👍🏻 looks good

mapkon:
Thanks @rwstauner This has been merged - will be part of the next release that we should ideally cut soonish.

PR #739, opened 2021-10-22
add proxy support for browser provider

alisade:
adding support for using a proxy for browser provider using ALL_PROXY environment variable (curl supported variable) - which supports both http and socks5 proxies


alisade:
adds support for passing a proxy server for browser auth, add support for custom sts endpoint url

PR #722, opened 2021-09-06
Fix SecTrustedApplicationCreateFromPath deprecated

peterkowalski:
Fixes:
```
cgo-gcc-prolog:203:11: warning: 'SecTrustedApplicationCreateFromPath' is deprecated: first deprecated in macOS 10.15 - No longer supported [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecTrustedApplication.h:59:10: note: 'SecTrustedApplicationCreateFromPath' has been explicitly marked deprecated here
```

Based on: https://github.com/buildkite/cli/pull/101

Issue #714, opened 2021-08-10
Okta not remembering the device, causing infinite loop


resnikb:
Hi,
I use Okta with Okta push. When I have sessions enabled, the first call goes through the whole workflow and saves the session cookie. However, the next call causes saml2aws to hang, in an infinite loop. Using `--verbose` parameter shows that it is continuously cycling through `authWithSession` -> `follow` -> `Authenticate` -> `authWithSession` ...

I debugged the application, and the main issue seems to be that the call to `verifyMfa` passes `rememberDevice` property only in the body of the request - however, according to [Okta documentation](https://developer.okta.com/docs/reference/api/authn/#verify-push-factor), this should be a query parameter.  The existing code works when the verification link in the Okta response contains `rememberDevice` parameter. However, this was not always the case for me (I'm not sure why).

I have verified this by modifying the `oktaVerify` code `getMfaChallengeContext` method to do
```
    if !strings.Contains(oktaVerify, "rememberDevice") {
        separator := "?"
        if strings.Contains(oktaVerify, "?") {
            separator = "&"
        }
        oktaVerify = oktaVerify + separator + "rememberDevice=" + strconv.FormatBool(oc.rememberDevice)
    }
```

After this change, the device is remembered and the infinite loop doesn't happen.

I'm happy to submit a PR with the above change - but please be warned that I'm a complete Go newbie.

resnikb:
On further experimentation, there is another issue - but this one happens when the Okta push needs to be activated again (after 24h in my case). When that happens, the code path in the `follow` method which checks which response was received, falls into the `else` branch and initiates another call to `loginDetails.URL` - however, this time it does it as a new request and does not preserve any existing cookies - in particular, the Okta session cookie is not used.

The response received here explicitly clears the sesion cookie in the cookie jar, which then causes an infinite loop again.
If the session cookie is preserved in this call, then it is passed on as part of primary authentication, and I receive Okta push - and `saml2aws` works fine from there.

In my workaround, I just replaced the call to `http.NewRequest` with in-place modification of `req` - but I don't know if that is the right thing to do:
```
		req.URL, err = url.Parse(loginDetails.URL)
		if err != nil {
		  return "", errors.Wrap(err, "error building app request")
		}
		req.Method = "GET"
		req.Body = nil
```

cwalkatron:
[...]

PR #740, opened 2021-10-23
Share Okta credentials cross accounts

tommywo:
Allow splitting Okta login url to url and app_id so user can login once to okta and then use the same session to access different accounts, similarly like they can in the browser.

Configuration example:
```
[account-a]
app_id           = /app/example/AAAAAAA/sso/saml
url              = https://example.okta.com
provider         = Okta
(...)

[account-b]
app_id           = /app/example/BBBBBBB/sso/saml
url              = https://example.okta.com
provider         = Okta
(...)
```
The change from the SecClassInternetPassword to SecClassGenericPassword is because on the first login saml2aws would first save the okta session cookie to keychain and then delete it while saving the password. It also returned the token instead of password. this doesn't happen for me with SecClassGenericPassword.

I tested the change combined with pull request #722 and the fixes from issue #714


tommywo:
@wolfeidau What do you think about this pull request?

wolfeidau:
@tommywo changing the sec class would be a breaking change as keychain doesn't really handle this well... 

We need to come up with a way to enable this without breaking existing users.

briantist:
This would be a really great feature for us too, I hope to see it accepted!

PR #744, opened 2021-11-19
Adding Pinentry as a prompter

sledigabel:
This PR adds a new type of Prompter called `Pinentry`.

## Why?
The current prompter is very convenient for many things, but it requires a TTY at all time. Which is mostly the case when you use it and require the prompter, like a first time login, or selecting a role in the list.
However, when using `saml2aws` with the credential process, it only works if there is no 2FA or if the 2FA is detached (typically, using an app like MSAUTH). For users having to type their TOTP, the TTY is required.
Unfortunately, the AWS SDK is making calling the credential process and doesn't allow STDIN, hindering the ability to automate the TOTP. There _is_ a `--mfa_token` option, but requires some additional wrapping behind it to get the TOTP in real time and doesn't account for when the credentials are already valid or when the 2FA token is still cached.

## How?
I found that the `Prompter` ([link](https://github.com/Versent/saml2aws/blob/master/pkg/prompter/prompter.go#L6-L13)) type is an interface, which is currently implemented only by `survey` ([link](https://github.com/Versent/saml2aws/blob/master/pkg/prompter/survey.go#L11-L12)), and thought I'd try to give a try to detach fairly generically the TOTP passcode prompt.
[Pinentry](https://github.com/GPGTools/pinentry) is a utility that ships with gpg as a way to enter safely a passphrase. It's common on every platform, it's been stable for a long time, and most OS ship with a window-based version of pinentry. Typically the one for mac can be compiled via brew: `brew install pinentry-mac`.

The change adds a new prompter that calls the old one in the background, and only for `RequestSecurityCode` would run pinentry to get the MFA TOTP token back.
This is an example using `pinentry-mac`
image

Now saml2aws can be used as a credential process for TOTP users too.

## How to use it?
There is a new parameter called `prompter` available in the common options.
You can either override it when running your usual commands, or set it once during the `configure` stage.
```console
brew install pinentry-mac
saml2aws configure --prompter=pinentry-mac [...]
```

At the moment, the accepted values for the `prompter` flag are:
- "", "default", "survey" (the existing behaviour)
- "pinentry" (your default pinentry program), "pinentry-_something_" (pinentry-mac, pinentry-gnome3)

### DIFF
--- a/cmd/saml2aws/main.go
+++ b/cmd/saml2aws/main.go
@@ -82,6 +82,7 @@ func main() {
 	app.Flag("session-duration", "The duration of your AWS Session. (env: SAML2AWS_SESSION_DURATION)").Envar("SAML2AWS_SESSION_DURATION").IntVar(&commonFlags.SessionDuration)
 	app.Flag("disable-keychain", "Do not use keychain at all. This will also disable Okta sessions & remembering MFA device. (env: SAML2AWS_DISABLE_KEYCHAIN)").Envar("SAML2AWS_DISABLE_KEYCHAIN").BoolVar(&commonFlags.DisableKeychain)
 	app.Flag("region", "AWS region to use for API requests, e.g. us-east-1, us-gov-west-1, cn-north-1 (env: SAML2AWS_REGION)").Envar("SAML2AWS_REGION").Short('r').StringVar(&commonFlags.Region)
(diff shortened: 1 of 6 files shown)


abualy:
facing same issue, thank you for the PR @sledigabel :D
:+1:

PR #745, opened 2021-11-20
Okta MFA - Present name of MFA. Prompt when multiple of same type are found

PerGon:
2 issues being addressed in here.

1. When multiple types of the same auth are found (e.g. multiple FIDO), it's not possible to identify which one to pick, given that in the prompt they are not possible to identify. See screenshot. 
image
This is fixed by adding the MFA name. See screenshot:
image

2. And most importantly, when the MFA options have the same name and the user is presented with a prompt, independently of the chosen option, the first one will always be the same picked. This is because the [survey code](https://github.com/Versent/saml2aws/blob/master/pkg/prompter/survey.go#L58-L62) will return the first index of any matched string (so 2 equal strings, and the same index is always returned, even if user picks 2nd). This is fixed by adding the auth name as pointed in the first point.

The above screenshots were taken when having MFA in "AUTO" mode. If the user has configured the MFA into a specific mode, FIDO for example, the following prompt is added indicating there's multiple FIDO options available:
image

### DIFF
--- a/pkg/provider/okta/okta.go
+++ b/pkg/provider/okta/okta.go
@@ -598,10 +598,13 @@ func getStateTokenFromOktaPageBody(responseBody string) (string, error) {
 	return strings.Replace(match[1], `\x2D`, "-", -1), nil
 }
 
-func parseMfaIdentifer(json string, arrayPosition int) string {
+func parseMfaIdentifer(json string, arrayPosition int) (string, string){
 	mfaProvider := gjson.Get(json, fmt.Sprintf("_embedded.factors.%d.provider", arrayPosition)).String()
 	factorType := strings.ToUpper(gjson.Get(json, fmt.Sprintf("_embedded.factors.%d.factorType", arrayPosition)).String())
-	return fmt.Sprintf("%s %s", mfaProvider, factorType)
+	// Okta gives names to some authentication methods
+	// displaying this name is useful when there's multiple auths of the same type. e.g. multiple FIDO options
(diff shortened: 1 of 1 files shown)


QuentinBtd (approved):

szechyjs:
@QuentinBtd @wolfeidau can we get this merged? this is a real issue that affects us daily.

pkg/provider/okta/okta.go
```diff
@@ -598,10 +598,13 @@ func getStateTokenFromOktaPageBody(responseBody string) (string, error) {
...
 
-func parseMfaIdentifer(json string, arrayPosition int) string {
+func parseMfaIdentifer(json string, arrayPosition int) (string, string){
 	mfaProvider := gjson.Get(json, fmt.Sprintf("_embedded.factors.%d.provider", arrayPosition)).String()
 	factorType := strings.ToUpper(gjson.Get(json, fmt.Sprintf("_embedded.factors.%d.factorType", arrayPosition)).String())
-	return fmt.Sprintf("%s %s", mfaProvider, factorType)
+	// Okta gives names to some authentication methods
+	// displaying this name is useful when there's multiple auths of the same type. e.g. multiple FIDO options
+	authName := gjson.Get(json, fmt.Sprintf("_embedded.factors.%d.profile.authenticatorName", arrayPosition)).String()
+	return fmt.Sprintf("%s %s", mfaProvider, factorType), fmt.Sprintf("%s", authName)
```
rarruda, line 607: Might make the linter happier:
```suggestion
	return fmt.Sprintf("%s %s", mfaProvider, factorType), authName
```

PR #746, opened 2021-11-25
Add the time when it will expire to debug log

mikutas:
Do we want to know when to expire even if credentials are not expired for now ?


pkg/awsconfig/awsconfig.go
```diff
@@ -113,13 +113,13 @@ func (p *CredentialsProvider) Load() (*AWSCredentials, error) {
 }
 
 // Expired checks if the current credentials are expired
-func (p *CredentialsProvider) Expired() bool {
+func (p *CredentialsProvider) Expired() (bool, *time.Time) {
```
sledigabel, line 116: The function name doesn't really do what it says now.
Why not create a new method `ExpiresAt` to return the `*time.Time` instead?

pkg/awsconfig/awsconfig.go
```diff
@@ -122,6 +122,15 @@ func (p *CredentialsProvider) Expired() bool {
 	return time.Now().After(creds.Expires)
 }
 
+func (p *CredentialsProvider) ExpiredAt() *time.Time {
```
sledigabel, line 125: I would recommend passing it as value, since a `time.Time` is a measure of Time.

sledigabel (commented):
Looks nice 👍  just a small comment on using value semantic

pkg/awsconfig/awsconfig.go
mikutas: Just noticed :eyes:, can we perhaps use previousCreds.Expires instead of adding ExpiredAt?

sledigabel: @mikutas that is indeed a _lot_ simpler :-)

sledigabel:
FWIW, LGTM! 🚀

gliptak:
@mapkon please rebase

mapkon:
> @mapkon please rebase

Do you want me to merge this?

gliptak:
yes please consiider merging. @mikutas found this useful for debugging

PR #753, opened 2021-12-21
Add support for skipping parameter prompt via environment variable

marcus-crane:
This PR adds support for automatically invoking the `--skip-prompt` flag by setting `SAML2AWS_SKIP_PROMPT=true`.

I've tried this locally and it works as expected so thankfully, it's a change with a very small scope!

I use OneLogin with `saml2aws`, along with my credentials stored using macOS keychain. It works great as a `credential_process` for `awscli` but when running normally ie; `saml2aws login -a "blah"`, it can be a pain to be asked to confirm my credentials each time, since I've already saved them.

Internally we have some shared environment variables (ie; a shell script you can import) so it'll be nice to just set this one for all developers and have it work rather than say; aliasing `saml2aws` to include the `--skip-prompt` flag or remembering to specify it each time.

Thanks!


kylesloan:
Would love to see this get merged in.

PR #763, opened 2022-01-28
helper/linuxkeyring: use configured pass directory

### LINKED ISSUE #440: pass keychain doesn't use correct path
I am using the default pass keychain for Linux. I have an environment variable set `PASSWORD_STORE_DIR` but `saml2aws` doesn't recognize it when configuring my user. I have to unset it and initialize a new password store at the default path.

eonpatapon:
When the keyring package is configured and PassDir is empty it will
default to ~/.password-store/.

See https://github.com/99designs/keyring/blob/master/pass.go#L26

Closes #440

### DIFF
--- a/helper/linuxkeyring/linuxkeyring.go
+++ b/helper/linuxkeyring/linuxkeyring.go
@@ -2,6 +2,7 @@ package linuxkeyring
 
 import (
 	"encoding/json"
+	"os"
 
 	"github.com/99designs/keyring"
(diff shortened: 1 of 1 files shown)


tinaboyce:
hi @eonpatapon , is this PR still relevant? I've noticed the issue this PR meant to close on merge is already closed.

eonpatapon:
By looking at the current code I believe the issue still exists

Issue #728, opened 2021-09-20
Is google prompt with extra number challenge supported?


miguelaferreira:
I'm getting an extra numeric challenge, when authenticating using Google IdP and Google Prompt as the 2FA mechanism.
However, `saml2aws` is not giving me the number I need to select on my phone.

Is this extra challenge supported?

See this screenshot with the challenge has it happens in the browser:
![image](https://user-images.githubusercontent.com/4670993/134012748-bf5097a1-5ff0-4fdb-9bb1-02c4cd37bdfe.jpg)

Enviii:
Do Google prompts work for you? I (and another colleague) have not received any today when attempting to authenticate.

miguelaferreira:
Yes, I'm getting the prompt.

msmenegol:
I've been having this issue as well. The challenge appears on my phone after I answer the prompt, but no number get displayed on the terminal. I tried running saml2aws with ```--verbose```, but it seems that the solution doesn't get printed anywhere as well.
![2fa_challenge](https://user-images.githubusercontent.com/23125726/134186494-07801db5-6f97-49da-aeca-8ffbc9cb4b71.jpg)

Issue #477, opened 2020-04-08
new type of verification when using Google app prompt for MFA


gilmoregrills:
This issue wasn't found by me, rather by @tobystaines, but when using saml2aws to log in with his GSuite credentials he was presented with a new type of verification screen that I've not seen before: 

![IMG_0228](https://user-images.githubusercontent.com/11209344/78801798-03eb1180-79b5-11ea-9e44-d70a14a7c19a.PNG)

saml2aws didn't return any number in his terminal, which is unsurprising assuming this kind of verification step is super new. Anyone else seen this?

FernandoMiguel:
I had open a similar one for Microsoft Authenticator 
https://github.com/Versent/saml2aws/issues/473

testlnord:
I've faced a similar issue.

PR #777, opened 2022-02-22
feat(googleapps): add support for 'extra number' in device push challenges

nickvollmar:
This PR addresses issues https://github.com/Versent/saml2aws/issues/477 and https://github.com/Versent/saml2aws/issues/728.

In some cases, Google's device push challenges come with an 'extra number', which the user must enter on their phone for the prompt to succeed. Personally, I get an 'extra number' whenever I attempt to sign in from a new WiFi network. (To reproduce, you can use a VPN and log in from a country that you've never been to.)

#### Example output with 'extra number' challenge
```
Using IdP Account default to access GoogleApps https://accounts.google.com/o/saml2/initsso?...
Authenticating as nick@example.com ...
Check your phone and tap 'Yes' on the prompt, then tap the number:
	46
Then press ENTER to continue.

Selected role: arn:aws:iam::123456789012:role/foobar
Requesting AWS credentials using SAML assertion.
Logged in as: arn:aws:sts::123456789012:assumed-role/foobar/nick@example.com
```

#### Example output without extra number
```
Using IdP Account default to access GoogleApps https://accounts.google.com/o/saml2/initsso?...
Authenticating as nick@example.com ...
Check your phone and tap 'Yes' on the prompt. Then press ENTER to continue.

Selected role: arn:aws:iam::123456789012:role/foobar
Requesting AWS credentials using SAML assertion.
Logged in as: arn:aws:sts::123456789012:assumed-role/foobar/nick@example.com
```

The jsname attribute is being used elsewhere in this file (`func extractKeyHandles`) so I hope it's safe to use here too.


alsaghir:
Is there any idea when to merge this ?

miguelaferreira:
I would also very much like this feature to be merged and releases. My team has had to disable google prompt altogether since we reported #728.

TWRogers:
Seconded

nickvollmar:
Hello @wolfeidau, I'm sorry to bother you –– but I notice you've merged some PRs lately!

May I ask for review on this change?

I will happily sign any contributor/copyright agreement if Versent needs that.

martin-shippit:
I've tried this PR and looks like it solves the issue. Nice work. My only concern would be the lack of tests. Perhaps it's not feasible :thinking:

martin-shippit (approved):

nickvollmar:
Hello @martin-shippit, sorry for the delay. I have added a test and merged the latest changes from upstream. Could you re-approve and perhaps merge?

martin-shippit:
Beautiful. I don't have merge privilege but this looks good to me. I've been using this version for a while and it works well.

@wolfeidau could you review and merge if it looks good to you?

jmctune:
@wolfeidau :wave: Several users in my org get stuck with this exact use case and have to either guess the correct number or hope Google presents them with a different challenge. Could this be reviewed / merged?

nickvollmar:
Thank you @mapkon! I believe I have addressed the linter errors, could you please re-run the GitHub action?

mapkon:
> Thank you @mapkon! I believe I have addressed the linter errors, could you please re-run the GitHub action?

Superb! I will take a look - please watch out for the code coverage, incase it regresses.

pkg/provider/googleapps/googleapps.go
```diff
@@ -737,3 +744,11 @@ func isAppId(val string) string {
 	}
 	return appId
 }
+
+func extractDevicePushExtraNumber(doc *goquery.Document) string {
+	extraNumber := ""
+	doc.Find("div[jsname=feLNVc]").Each(func(_ int, s *goquery.Selection) {
```
mapkon, line 750: I am wondering if it is possible for this to change? It looks like a generated name, which might  be different from page to page?

nickvollmar: I agree that this is a generated name, and therefore it's less than ideal to use. But it seems to me that this is the best we can do. 

I checked in an anonymized version of the HTML at `pkg/provider/googleapps/example/challenge-extra-number.html` –– as you can see, if we don't want to use generated names, we have to couple ourselves to the HTML structure, which is not ideal either.

A couple of mitigating circumstances that make me feel better about the usage of jsname:
1. Elsewhere in this file we have `func extractKeyHandles` that relies on a jsname as well.
2. [@martin-shippit's comment on Aug 31 '22](https://github.com/Versent/saml2aws/pull/777#issuecomment-1233520239) indicates that this logic worked for him too, so his page must have had the same jsname.

mapkon: OK, thanks for the explanation.

mapkon (approved):

Issue #120, opened 2017-12-29
Support credential_process feature of CLI


hoegertn:
It would be great to support: http://docs.aws.amazon.com/cli/latest/topic/config-vars.html#sourcing-credentials-from-external-processes

I think #109 and #55 need to be implemented before this and caching of the temp credentials has to work.

This would be the best integration into the CLI I think.

fdamstra:
Another vote for this. It would be nice to have authentication happen transparently through the `credential_process` directive in `~/.aws/credentials`.

Some additional information is here: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html

To support this, the script needs to generate output of the following format:
```
{
  "Version": 1,
  "AccessKeyId": "an AWS access key",
  "SecretAccessKey": "your AWS secret access key",
  "SessionToken": "the AWS session token for temporary credentials", 
  "Expiration": "ISO8601 timestamp when the credentials expire"
}  
```

However, for this to be truly useful, saml2aws would also need to cache successful logins (e.g. #492).

shousper:
I ended up writing a wrapper that essentially looks like this:

```bash
#!/usr/bin/env bash

# Read AWS profile, role name and role ARN from parameters
profile=$1
role=$2
role_arn=$3
region=${4:-ap-southeast-2}

# Use cache credentials file
export AWS_SHARED_CREDENTIALS_FILE=~/.saml2aws-credential-cache

# Save credentials to cache profile
saml2aws login --skip-prompt --region "$region" --profile "${profile}-${role}" --role "$role_arn" >&2

# Read in credentials to reformat as JSON
eval "$(saml2aws script --profile "${profile}-${role}")"

# Render JSON required by AWS CLI
echo '{"Version": 1}' |
	jq --arg v "$AWS_ACCESS_KEY_ID" '.AccessKeyId = $v' |
	jq --arg v "$AWS_SECRET_ACCESS_KEY" '.SecretAccessKey = $v' |
	jq --arg v "$AWS_SESSION_TOKEN" '.SessionToken = $v' |
	jq --arg v "$AWS_CREDENTIAL_EXPIRATION" '.Expiration = $v' |
	jq -c '.'
```

You cannot have the static credentials in `~/.aws/credentials` AND set `credential_process` so the use of an external file was required.

The `~/.aws/config` was then populated with profiles like this:

```
[profile uat-blah]
region = ap-southeast-2
[...]

PR #790, opened 2022-03-14
Use prompter.RequestSecurityCode for MFA in OneLogin

patbos:
Relates to #120.
With this fix it is possible to use credential_process with OneLogin with MFA requirement.


patbos:
Any plans of merging this or should I close it?

PR #796, opened 2022-03-26
Use RequestSecurityCode prompt function to ask for MFA in Okta provider

FaHeymann:
https://github.com/Versent/saml2aws/pull/744 introduced pinentry as an alternative prompter to allow the user to enter a TOTP token while using saml2aws as the credential_process. To make this work in combination with the Okta provider I would like to change the prompt for MFA in there to also use the `RequestSecurityCode` prompt function.


christianmeyer:
duplicate of #790

FaHeymann:
It's a similar change but in a different provider

Issue #815, opened 2022-05-02
[Browser Provider] Support browser type


eliat123:
Whenever using the `Browser` idp-provider, the default launched browser is `Chromium,` which the `playwright-go` package supports by default.

This behaviour is acceptable in most cases; however, when there are specific self-signed trust certificates, the default `Chromium` does not recognize installed certificates and already trusted CAs, and there is a prompt to approve them.

To fix it, we would like to add to the `saml2aws` configuration a `browser_type` property for the `Browser` Idp-provider, which will set it in the playwright launch configuration.

MatteoJoliveau:
Cross-linking my comment on the relevant PR: https://github.com/Versent/saml2aws/pull/816#issuecomment-1174781406

This feature would allow `saml2aws` to work flawlessly on NixOS/other non-FHS linux distributions

merusso:
Personally, I'd love it if this feature would use the system's default browser instead of downloading a playwright browser, if that's possible.

merusso:
> Cross-linking my comment on the relevant PR: [#816 (comment)](https://github.com/Versent/saml2aws/pull/816#issuecomment-1174781406)
> 
> This feature would allow `saml2aws` to work flawlessly on NixOS/other non-FHS linux distributions, which are broken at the moment.

@MatteoJoliveau, I'm not sure about this statement. I asked for clarification on the PR.

eliat123:
Closing as https://github.com/Versent/saml2aws/pull/816 was merged

Issue #729, opened 2021-09-23
Using locally installed browser rather than outdated sanboxed chromium from "playwright-go"


thom-vend:
Hi
I'm starting **loving** this tool however there is one little part which isn't great: the sandboxed browser.

1) First the sandboxed chromium run an **outdated** version and indeed **vulnerable**: `90.0.4430.0` on my mac.  

_I've installed sam2aws 1~2 months ago and only run `saml2aws login --idp-account "$aws_app" --profile "$aws_app" --skip-prompt --cache-saml` when my AWS tmp credential are outdated or when I'm assuming another role/other account. saml2aws logs says `Downloaded browsers successfully` so it make me think it try to update this chromium?_ 🤔 

2) Using a private sandboxed chromium force us to re-login each time. 

_My main browser is already connected to the saml identity provider, users are wasting a bit their time (Ex when using multiple aws account and roles)_

**How could we use a locally installed browser?** 🤔 

_Or not using a browser at all with provider like onelogin?_

Thanks for you engagement in this nice tool,
Cheers,
Thom

renanrt:
I would love that too!

fllaca:
I also think this can be a good feature, the tool can open a link in your browser in where you can supply credentials and go back to the CLI.

merusso:
In my opinion, using the system's default browser makes more sense than downloading/caching a new browser binary for login.

richard-pianka:
This would be a huge improvement, especially given the recent issues with using Google Apps as the identity provider.

GuillaumeRoss:
This would be an important features for companies that wish to restrict access to AWS only to corporate machines. For example, using Google's Context-Aware Access (CAA) feature allows one to only let a SAML app work from company owned machines (as well as checking a few other characteristics of the client machine), but it requires using a Chromium/Chrome session with the Endpoint Verification extension, logged in to Google Workspace.

If saml2aws supported using existing Chrome browsers, this would instantly work. Alternatively, there would have to be a way to orchestrate downloading playright, the extension, and ensuring the user logs into Google Workspace at the browser level, making the extension sync, which sounds a lot more brittle.

sahaqaa:
I would love see this implemented - because each time entering credentials when i need to switch between accounts / roles in multi-account AWS environment is painful :-(

kbarlowgw:
Another +1 here.

dannysauer:
[...]

PR #816, opened 2022-05-03
Added support for browser-type and browser-executable-path

eliat123:
### Improvement details:
This is the implementation to https://github.com/Versent/saml2aws/issues/815 

An additional `saml2aws` command line argument & configuration to allow to the specification of `BrowserTypeLaunchOptions.Channel` for 'Browser' idp-provider.

### Usage example:
- This allows opening a chrome browser (instead of Chromium) when running saml2aws.
```
$ saml2aws configure --idp-provider=Browser --browser-type=chrome --username=${USERNAME} --region=${REGION} --url=${URL} --role=${AWS_ROLE} --idp-account=${IDP_ACCOUNT} --profile=${IDP_ACCOUNT} --skip-prompt
$ ${SAML2AWS} login --idp-account=${IDP_ACCOUNT} --profile=${IDP_ACCOUNT} --skip-prompt
```


eliat123:
@wolfeidau If possible, can I kindly ask you to have a look at this addition to the 'Browser' provider and let me know what you think.

efernandes-dev-ops:
@eliat123  I believe this might also resolve [729](https://github.com/Versent/saml2aws/issues/729).

MatteoJoliveau:
Hi, I would like to push this PR because it would allow NixOS users like me to use the Browser provider. In NixOS you cannot simply download a random executable from the internet and run it because it doesn't use the Linux Filesystem Hierarchy Standard (FHS), therefore the setup script and subsequently downloaded chromium (after I patched the script by hand) won't work (see screenshots)

image
image

As a rule of thumb downloading executables to the user machine is not a really good practice. I prefer having complete control over what is in my system and installing software through the package manager.

The other thing that bothers me with the downloaded chrome executable is that it is run in Incognito mode, meaning I have to re-authenticate to my Google account every time. If saml2aws were to use the system browser (firefox in my case) it could reuse the active login session.

mlai:
Looks like a good feature. Can you help to review and merge?

efernandes-dev-ops:
@MatteoJoliveau We have the same usecase with regards to the chrome executable running in incognito mode which is a blocker for us in adopting this atm.

eliat123:
@wolfeidau kind reminder.
If possible, can I kindly ask you to have a look at this addition to the 'Browser' provider and let me know what you think?

eliat123:
Thank you @TwiN.
As requested, I pulled the recent changes from the upstream branch so that @wolfeidau can review & merge it in one go when they have time.

cmd/saml2aws/main.go
```diff
@@ -70,6 +70,7 @@ func main() {
 	app.Flag("config", "Path/filename of saml2aws config file (env: SAML2AWS_CONFIGFILE)").Envar("SAML2AWS_CONFIGFILE").StringVar(&commonFlags.ConfigFile)
 	app.Flag("idp-account", "The name of the configured IDP account. (env: SAML2AWS_IDP_ACCOUNT)").Envar("SAML2AWS_IDP_ACCOUNT").Short('a').Default("default").StringVar(&commonFlags.IdpAccount)
 	app.Flag("idp-provider", "The configured IDP provider. (env: SAML2AWS_IDP_PROVIDER)").Envar("SAML2AWS_IDP_PROVIDER").EnumVar(&commonFlags.IdpProvider, "Akamai", "AzureAD", "ADFS", "ADFS2", "Browser", "GoogleApps", "Ping", "JumpCloud", "Okta", "OneLogin", "PSU", "KeyCloak", "F5APM", "Shibboleth", "ShibbolethECP", "NetIQ", "Auth0")
+	app.Flag("browser-type", "The configured browser type when the IDP provider is set to Browser. (env: SAML2AWS_BROWSER_TYPE)").Envar("SAML2AWS_BROWSER_TYPE").EnumVar(&commonFlags.BrowserType, "chrome", "chrome-beta", "chrome-dev", "chrome-canary", "msedge", "msedge-beta", "msedge-dev", "msedge-canary")
```
TwiN, line 73: Is there a reason why there's no support for Firefox? 👀

zemberdotnet:
```go
browser, err := pw.Chromium.Launch(launchOptions)
	if err != nil {
		return "", err
	}
```
It looks like it [always launches a Chromium browser](https://github.com/Versent/saml2aws/blob/f45c5f888cfdf9dcd1e8863e2301ff957e03d787/pkg/provider/browser/browser.go#L46-L49), so I'd guess that's the reason why.

eliat123: @TwiN, @zemberdotnet Added Firefox.

The original values were set based on the comment of the supported values by the playwright Channel; see: https://github.com/playwright-community/playwright-go/blob/8e8f670b5fa7ba5365ae4bfc123fea4aac359763/generated-structs.go#L441-L444

However, it totally made sense to me to add this capability, so I added support for Firefox and Webkit. (`msedge` is also supported using Chromium which is the default).

eliat123: @zemberdotnet thank you for noticing that.

The reason for this behaviour is that the `commonFlags` can be passed both in the `configure` or other steps such as `login`. In my case, i have set the `--browser-type` in the `configure` and then it was available in the login.

To support providing the flag in the login command, I removed the additional check for `commonFlags.IdpProvider == "Browser"` since the BrowserType is only used by the `Browser` Provider, the functionality is still the same.

Please check it again as it should work now.

merusso:
> Hi, I would like to push this PR because it would allow NixOS users like me to use the Browser provider. In NixOS you cannot simply download a random executable from the internet and run it because it doesn't use the Linux Filesystem Hierarchy Standard (FHS), therefore the setup script and subsequently downloaded chromium (after I patched the script by hand) won't work (see screenshots)

@MatteoJoliveau, does this PR fix the issue you described? Skimming the code, I'm thinking it still uses a [playwright](https://github.com/playwright-community/playwright-go) browser, something that was downloaded to the playwright cache directory. It just allows users to change the playwright browser being used instead of only using Chromium.

MadOx710:
Definitely looking forward to Provider Browser support outside of Chromium.  Context-Aware policies in Google Workspace leverages the Endpoint Verification Chrome plugin, and currently no Provider option for saml2aws works when Context-Aware policies are enabled.

eliat123:
@MadOx710, @MatteoJoliveau, @zemberdotnet, @TwiN, looking at the previous comments, are we ok with the [playwright](https://github.com/playwright-community/playwright-go) browser for now? or do want to wait for non-downloadable browser support?
If we are ok to continue with the first approach, let's merge this PR. 
Otherwise, is there a feature request for non-downloadable browser support? Do you have a GitHub issue with its requirements? Is someone actively working on it? 

I would be happy to contribute, but I would like to know the direction of this project.

mapkon:
@eliat123 

@gliptak and I did some work on browser.go! It has caused conflicts with this PR. Please take a look and let me know if this is still required.

gliptak:
@eliat123 please rebase

relatedly, would https://github.com/Versent/saml2aws/pull/816/files#diff-5cd59a5a8d31c61bb9c035ee3524699cd8e97b722683501d805074f7d11b7ee9R45-R50 all work for all OS types?

GuillaumeRoss:
> @MadOx710, @MatteoJoliveau, @zemberdotnet, @TwiN, looking at the previous comments, are we ok with the [playwright](https://github.com/playwright-community/playwright-go) browser for now? or do want to wait for non-downloadable browser support? 

I just commented in https://github.com/Versent/saml2aws/issues/729#issuecomment-1591152642 - essentially the ability to use the "real" browsers installed on a system provides significant security benefits not only on the client side (not having another browser executable+drivers downloaded and executed) but on the cloud side (ensuring only trusted clients can connect to AWS). I think the requirements fit this other issue quite well and so that discussion doesn't have to live in his one here.

eliat123:
@GuillaumeRoss playwright-go comes with "chromium," "firefox", "webkit", see: https://playwright-community.github.io/playwright-go/

This means that If you don't have one of the three browsers installed, it will download the browser for you and then perform the browser IDP authentication.

I have also added the ability to set the `--browser-executable-path`. 
This means that playwright-go will execute the browser at your local path instead of downloading the browser, this means that you can configure to use the browser of your choice.

In addition, playwright-go supports a set of predefined browser types that are not downloadable, this means that if you have already installed that browser and set the `--browser-type` to it it will use the installed browser.
The list for that is one of the predefined list:
```
"chrome", "chrome-beta", "chrome-dev", "chrome-canary", "msedge", "msedge-beta", "msedge-dev", "msedge-canary"
``` 

I have tested this with `chrome`, `chrome-canary`, `msedge` and verified that it uses the `ExecutablePath` instead of the default one.

I believe this resolves most of the issues in https://github.com/Versent/saml2aws/issues/729 

@mapkon @gliptak please review

mapkon (commented):
This looks on. Does that mean that we should close #815?

mapkon (changes requested):
Please checkout the linting errors and fix them.

eliat123:
> Please checkout the linting errors and fix them.

@mapkon Thank you for noticing that, I updated the golint version to be able to see the ci lint issue and fixed it.

dangymbol:
Long comment:  
This looks like a big improvement for any nix environment.  I'm using Linux Mint which Playwright doesn't currently support due to incorrect detection of the base linux version.  That means running aws2saml with the browser provider gives me a delightful segmentation violation!  This PR should fix that by allowing me to use the installed version instead.

I'm also not in favour of downloading binary versions of Chromium, Firefox and Webkit to a cache - since I already have the current ones installed through the standard package manager.

Short comment: +1 😄

mapkon:
@dangymbol Would you mind taking on this PR and updating it?

eliat123:
@mapkon I have updated this PR.
@dangymbol FYI.

mapkon (approved):

pkg/provider/browser/browser.go
```diff
@@ -56,10 +72,37 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 		Headless: playwright.Bool(cl.Headless),
 	}
 
-	// currently using Chromium as it is widely supported for Identity providers
+	validBrowserTypes := []string{"chromium", "firefox", "webkit", "chrome", "chrome-beta", "chrome-dev", "chrome-canary", "msedge", "msedge-beta", "msedge-dev", "msedge-canary"}
```
gliptak, line 75: why differentiate all these `browsertypes` if playwright only supports firefox, chrome and webkit?

eliat123: Hi @gliptak, Thank you for your review.

Playwright does support all of those browser types, see:
https://playwright.dev/docs/browsers and https://github.com/playwright-community/playwright-go/blob/main/generated-structs.go#L691C4-L695

The main idea behind this PR is that the user can choose a browser type which is already installed on their system.

If you choose a browser you do not have on your system, you will get a relevant error message.

For example, I tried to set `--browser-type=msedge`. Since I did not have `msedge` installed I got the following notification:  

```
INFO[0000] Setting browser type: msedge                  provider=browser
Error authenticating to IdP.: could not send message: Chromium distribution 'msedge' is not found at /Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge
Run "npx playwright install msedge"
```
After I installed `msedge`, it used the installed one:
```
INFO[0000] Setting browser type: msedge                  provider=browser
INFO[0001] opening browser                               URL="[IDP_URL]" provider=browser
INFO[0021] waiting ...                                   provider=browser
INFO[0049] clean up browser                              provider=browser
```

Furthermore, If your browser is installed on a non-default location, you can utilize `--browser-executable-path`, and it will start the browser from that path.

Hope this explains it.

pkg/provider/browser/browser.go
```diff
@@ -56,10 +72,37 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
...
+
+	if cl.BrowserType != "" {
+		logger.Info(fmt.Sprintf("Setting browser type: %s", cl.BrowserType))
+		launchOptions.Channel = playwright.String(cl.BrowserType)
+	}
+
+	// Default browser is Chromium as it is widely supported for Identity providers,
+	// It can also be set to the other playwright browsers: Firefox and WebKit
+	browserType := pw.Chromium
+	if cl.BrowserType == "firefox" {
```
gliptak, line 88: @eliat123 would there be more browsertypes mapped here?

https://playwright.dev/docs/browsers

eliat123: @gliptak Nope. Only these browsertypes are supported.

Ref: https://github.com/playwright-community/playwright-go/blob/ddc7abdd06db016d63664de1ab08870195085ddb/playwright.go#L21-L23

According to https://playwright.dev/docs/browsers :

1. Chromium is for "chromium", "chrome", "msedge", "chrome-beta", "msedge-beta" or "msedge-dev".
2. Firefox is for the recent [Firefox Stable](https://www.mozilla.org/en-US/firefox/new/) build.
3. Webkit is for the recent WebKit build.

Chromium will be the default option, but if "webkit" or "firefox" are selected, they will be used instead. This ensures the most available options.

gliptak: will be browserType be pw.Chromium when cl.BrowserType == 'msedge' and is this as expected?

eliat123: @gliptak The answer is yes. 

Here is the section from Playwright docs about Google Chrome & Microsoft Edge:
```
While Playwright can download and use the recent Chromium build, 
it can operate against the branded Google Chrome and Microsoft Edge browsers available on the machine (note that Playwright doesn't install them by default). 
In particular, the current Playwright version will support Stable and Beta channels of these browsers.

Available channels are 'chrome', 'msedge', 'chrome-beta', 'msedge-beta' or 'msedge-dev'.
```
ref: https://playwright.dev/docs/browsers#google-chrome--microsoft-edge

gliptak (approved):

mikex99:
Can I get a link to this build for testing?

gliptak:
@mikex99 bottom of https://github.com/Versent/saml2aws/actions/runs/6807857396

merusso:
Does this implementation allow use of an system's existing, default browser (#729)? It seems like it enables use of other Playwright browsers, which are still managed separately.

kbarlowgw:
I am trying to use this to run a local browser not in incognito mode but, while I can run a local chromium install, it still seems that playwright wants to open in it incognito mode.  How do we prevent this?

kbarlowgw:
Following up on this to see if there was a workaround or playwright switch I could use, it seems this ability is still in the works with playwright and we may have to update to use `browser.launchPersistentContext()` instead of `launch()` when complete:

https://github.com/microsoft/playwright/issues/11833

If someone knows of another option, please let me know.

Thanks.

merusso:
@gliptak I believe this was already created, #729. I believe the issue was mistakenly closed when this PR was merged.

Issue #818, opened 2022-05-06
No Duo prompt/authentication failure for Shibboleth provider


chris-reeves:
We use the Shibboleth provider in our org. Three days ago (starting on the 3rd of May) some of our users started reporting that they were no longer receiving a Duo prompt and getting the following error from saml2aws instead:
> Failed to assume role, please check whether you are permitted to assume the given role for the AWS service: error parsing aws roles: missing Assertion element

Investigating further, we found that Duo was responding to `GET /frame/prompt?sid=...` with an error:
```
GET /frame/prompt?sid=<redacted> HTTP/1.1
Host: api-redacted.duosecurity.com
User-Agent: saml2aws/1.0 (linux amd64) Versent
Content-Type: application/x-www-form-urlencoded
Referer: https://api-redacted.duosecurity.com/frame/web/v1/auth?tx=<redacted>
Accept-Encoding: gzip, deflate
Connection: close

HTTP/1.1 200 OK
Server: Duo/1.0
Date: Fri, 06 May 2022 10:53:31 GMT
Content-Type: text/html; charset=UTF-8
Connection: close
Cache-Control: no-store
Pragma: no-cache
Referrer-Policy: strict-origin-when-cross-origin
P3P: CP="IDC DSP COR ADM DEVi TAIi PSA PSD IVAi IVDi CONi HIS OUR IND CNT"
Etag: W/"458be520d99a661963c7d892f96bdbf2323930e7"
Set-Cookie: _xsrf="<redacted>"; HttpOnly; Path=/; SameSite=None; Secure
Strict-Transport-Security: max-age=31536000
Content-Security-Policy: default-src 'self'; frame-src 'self' ; img-src 'self'  ; connect-src 'self'    
Content-Length: 1503

<!DOCTYPE html>
<!--[if IE 8]> <html lang="en" class="ie ie8"> <![endif]-->
<!--[if IE 9]> <html lang="en" class="ie ie9"> <![endif]-->
<!--[if gt IE 9]><!--> <html lang="en"> <!--<![endif]-->
<head>
<meta charset="utf-8">

<title>
Two-Factor Authentication
</title>
<!-- Import for all versions of IE, in case the standards mode is downgraded to IE8 -->
<!--[if IE]> <link rel="stylesheet" href="&#x2f;frame&#x2f;static&#x2f;css&#x2f;v3&#x2f;ie8.css&#x3f;v&#x3d;e9b26"> <![endif]-->

[...]

PR #819, opened 2022-05-06
Fix Duo authentication failure with Shibboleth provider

chris-reeves:
Duo now appears to require the Accept-Language header to be present, otherwise it will produce an error (see #818).

This PR adds an Accept-Language header to the request.


chris-reeves:
@wolfeidau Any chance of getting this PR merged in? Our users are currently running off my fork, and it would be great to get them back on official releases.

Issue #847, opened 2022-07-09
Handle malformed Kmsi form field in ADFS


dboitnot:
I've encountered a situation where the value for `Kmsi` is "true " (with trailing space). This leads to:

```
DEBU[0003] HTTP Req URL="https://<redacted>/adfs/ls/IdpInitiatedSignOn.aspx?loginToRp=urn%3Aamazon%3Awebservices&client-request-id=3b.." http=client method=POST
DEBU[0003] HTTP Res Status="500 Internal Server Error" http=client
unable to classify response from auth server
```

Apparently this value is being pulled directly from the server's response:

https://github.com/Versent/saml2aws/blob/df3f6cf6757279e92c1dfe7e0155f0a60ec68d6b/pkg/provider/adfs/adfs.go#L75

dboitnot:
It turns out that, at least in my case, it's not trailing white space but a second, empty value for `Kmsi` in the form:

```go
len(authForm["Kmsi"]) // == 2

christianmeyer:
Just out of curiosity, is that issue also persisting with the version out of #795 ? In that case I would adopt the referenced fix.

PR #848, opened 2022-07-09
fix gh-847: Fixes malformed Kmsi value in form

dboitnot:
Addresses #847 by trimming all Kmsi values and dropping any empty ones.


mapkon (commented):
I suppose this is fairly straightforward a change. However, that file has zero tests to back it up - what is our confidence level on this?

dboitnot:
> I suppose this is fairly straightforward a change. However, that file has zero tests to back it up - what is our confidence level on this?

All I can say is it definitely solved my problem and I think it makes sense to prune the empty values. To be honest I have not been using this project as my client has switched to AWS SSO.

PR #850, opened 2022-07-11
Add `bgresponse: js_enabled` to challenge post too.

jmctune:
`loadFirstPage` function sends `bgresponse: enabled` to form data before posting, but this also needs to be done during the challenge POST, or you'll receive the infamous "This browser does not support Javascript.." message.

Small note: I haven't touched `go` outside of this project, so I'm unsure if there's a better way to do this, but this worked for me.


mjhoffman65:
This works for me.  Thanks!

wolfeidau:
@mjhoffman65 these sort of follow up "works for me" are SUPER helpful BTW, I can't test everything with these integrations so thanks!

Issue #843, opened 2022-07-01
Azure AD Number Match MFA


dannykitchen:
Azure AD Auth SSO - Microsoft soon to enforce number match 
https://docs.microsoft.com/en-us/azure/active-directory/authentication/how-to-mfa-number-match

Currently not supported 
Can you advise if will be covered ?

ablackrw:
It appears that a patch has been proposed to implement this matching as #857, but the change just seems to be sitting there. We are looking to roll out number matching in our MFA environment, and this lack of support is forming an impediment to the roll out.

ablackrw:
As a followup on my last comment, the change proposed in #857 is for ADFS, not AAD. I've put forward a patch for AAD support as #921, though I've only had limited success testing due to unrelated issues.

edgranau:
any updates on support for this? Number matching got activated in my client's org and `s2a` stopped working for me :( https://petri.com/microsoft-authenticator-number-matching-by-default/

sammcj:
Confirming saml2aws is now broken due to this.

sammcj:
Seeing the fallout from this today highlights to me what a critical piece of software this is. I wish AWS would either support it - or provide an official alternative.

johnfilo-kmart:
Yes, the issue is with AAD not ADFS. The ADFS changes were already merged (#857)

ajohanssonwalder:
This has been resolved by https://github.com/Versent/saml2aws/pull/829, incorporated into latest release [v2.36.3](https://github.com/Versent/saml2aws/releases/tag/v2.36.3) 🚀

PR #857, opened 2022-07-21
Fix ADFS auth when using MFA with number matching

alexbumbacea:
Based on usage it looks like p#instruction is populated via javascript after page load. The entropy number is not present in the p#instructions element.

https://docs.microsoft.com/en-us/azure/active-directory/authentication/how-to-mfa-number-match

### DIFF
--- a/pkg/provider/adfs/adfs.go
+++ b/pkg/provider/adfs/adfs.go
@@ -135,7 +135,15 @@ func (ac *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 			doc.Find("input").Each(func(i int, s *goquery.Selection) {
 				updatePassthroughFormData(azureForm, s)
 			})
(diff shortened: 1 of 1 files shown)


alexbumbacea:
@wolfeidau Can you help get this merged? Once this feature will get out of beta, it will be enforced in a lot of orgs. So it's going to become even more important. Thanks

edgranau:
https://github.com/Versent/saml2aws/issues/843#issuecomment-1432309160

sammcj:
@versent anything we can do to help test this to get it merged ASAP?

sammcj (approved):

mapkon:
Is it possible to add a test?

mapkon (approved):
This looks OK! However, I would sleep better with a unit test added to buttress the change

PR #862, opened 2022-08-02
Use the SHELL env var to determine which shell exec will use

everops-mat:


mapkon:
Ref: https://github.com/Versent/saml2aws/pull/1007#issuecomment-1490144184

PR #880, opened 2022-08-29
Expand credentials-file path if '~/' prefix

lyoung-confluent:
When using `credential_process` in `~/.aws/config` the AWS SDKs are not consistent in how the credentials process is executed. For example, [in aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2/blob/b7d8e15425d2f86a0596e8d7db2e33bf382a21dd/credentials/processcreds/provider.go#L93) the command is wrapped via `sh -c "..."` however [in botocore it is not](https://github.com/boto/botocore/blob/7d4a3fc41928155a4a1eac04c02b68af64a3a59d/botocore/credentials.py#L1049-L1054). 

As a result, you cannot provide a `credentials-file` path that is rooted in `$HOME` or `~/` without adding an additional `sh -c "..."` wrapper in your config. This PR modifies the handling of the `credentials-file` flag to explicitly expand `~/` to the user's home directory if present via `homedir.Expand`.


mapkon:
This is failing @lyoung-confluent

gliptak:
the expansion code would look

```
	expanded, err := homedir.Expand(filename)
	if err != nil {
		return nil, err
	}
        return expanded, nil
```

mapkon:
> the expansion code would look
> 
> ```
> 	expanded, err := homedir.Expand(filename)
> 	if err != nil {
> 		return nil, err
> 	}
>         return expanded, nil
> ```

Yeah, but the code needs to be fixed! That or we close this PR and restart a new one.

gliptak:
already implemented at
https://github.com/Versent/saml2aws/blob/master/pkg/awsconfig/awsconfig.go#L185-L188

gliptak:
consider closing

PR #884, opened 2022-09-05
Use the same output for surveys as for logs.

### LINKED ISSUE #883: Prompt output always goes to stdout instead of to the log stream
Prompts for user input always go to stdout, even though the default non-Windows output stream is stderr.

This breaks the `script` command and the `login --credential-process` commands, since the prompts will be invisible to the user _and_ will get written to the output stream before the script or JSON is.

jkinkead:
Add a package variable to hold the survey output writer and a setter for it.

Add a helper to generate a survey option using the writer for stdout.

Initialize the survey output to be the same as for the log module.

---

Fixes #883 .


jkinkead:
One thing I forgot to note: The `survey` library can't just query [`log.Writer`](https://pkg.go.dev/log#Writer) because it requires a special type that include a file descriptor, and there is no way to get an fd from `log`.

mapkon (changes requested):
@jkinkead Merging this will bump our test coverage down. Could you please consider adding tests?

mapkon (commented):
Any chance this can be worked on? We need parity with the current test coverage to merge

mapkon:
> Any chance this can be worked on? We need parity with the current test coverage to merge

@jkinkead

malept (commented):
> Merging this will bump our test coverage down. Could you please consider adding tests?

I think that any tests that could be added in this PR would be primarily performative, instead of, for example, being used to spot regressions in future PRs. From what I can tell, the `survey` module is not conducive to test writing to begin with.

PR #888, opened 2022-09-21
saml2aws as credential plugin for eks cluster

d-kononov:
Ability to use `saml2aws` binary file as [credential plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins) for EKS cluster


sledigabel:
Have you thought of using the `aws-iam-authenticator` and use the aws cli with credential process to auth seemlessly?
```
- name: myCluster
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - eks
      - get-token
      - --cluster-name
      - mycluster
      - --profile
      - mycluster
      - --region
      - us-east-1
      command: aws
      env: null
```

sledigabel:
@d-kononov we use ADFS as IdP, backed by Azure. We do not use AWS IAM Users at all. You get a role in the end which you can use to auth in EKS with RoleBinding.

mapkon:
@d-kononov Are you still interested in this? If yes, please fix conflicts so we can merge.

PR #798, opened 2022-03-27
Fix build issues on OSX

duckfez:
Removes deprecated godownloader from Makefile, downloads goreleaser directly via "go install".

Fixes #797

PR #890, opened 2022-09-28
Add minimal Dockerfile and a goreleaser docker config

BenPhegan:
I was looking to introduce the creation of a minimal docker image to help with using this in automated CI flows, and developer workstations.

Not sure of the approach you use to release, so I am more than happy to discuss and modify to ensure that this suits your workflow.  Also, it has been a while since I worked with GoLang tooling, so please let me know if you would prefer this to be implemented another way.

This also includes a few fixes to the Makefile to ensure that `goreleaser` is available for targets that require it, as it did not function on my machine out of the box.


Makefile
```diff
@@ -39,6 +39,13 @@ build: $(BIN_DIR)/goreleaser
 	$(BIN_DIR)/goreleaser build --snapshot --rm-dist
 .PHONY: build
 
+release-local: $(BIN_DIR)/goreleaser
+	$(BIN_DIR)/goreleaser release --snapshot --rm-dist
+.PHONY: release-local
+
+$(BIN_DIR)/goreleaser:
+	@GOBIN=$(BIN_DIR) go install github.com/goreleaser/goreleaser@latest
```
BenPhegan, line 47: Note that there is a `GORELEASER_VERSION` variable at the top of the Makefile, however this appeared to be unused, and did not relate to a particular version that was accessible via a `go install` command.

BenPhegan: Just noted a distinct similarity between this change and https://github.com/Versent/saml2aws/pull/798/files.  Somewhere in the middle lies an accurate fix.  It should not be a `.PHONY`, but the version constraint seems to work.  Will update.

BenPhegan: @duckfez Just a heads up that I hit the same issue as you, and fixed it prior to realising that there was a similar PR up already.

BenPhegan:
@wolfeidau - let me know if you need anything else on this?  Happy to discuss the approach further.  I noticed that there are about 6 images on Docker Hub already that are providing saml2aws (all old to some degree).  Would prefer we could get an official one published as part of the release process so it can stay updated.

BenPhegan:
@wolfeidau - any updates on this one?

BenPhegan:
Bump

mapkon:
@BenPhegan Thank you for this. Can you rebase from master and fix the conflicts? I am happy to merge after that

gliptak:
relatedly what is the major difference between `go` and `release` GHA? `goreleaser/goreleaser-action` usage seems to be similar

https://github.com/Versent/saml2aws/tree/master/.github/workflows

BenPhegan:
@mapkon - this is a little more complicated now.  With the inclusion of this commit: https://github.com/Versent/saml2aws/commit/08a0a70fb1afc96211317560245c783fbe7d24aa it creates an issue with lightweight docker images, as it becomes difficult to statically compile and distribute a single file binary.  This seems to be due to taking on a dependency for hidraw, which brings in libudev, which according to [this](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=915566) is not built statically.

I could:

1. Change the `FROM` to a more complete Docker image, however that is going to make this a much larger download.   
2. Look at pulling in and compiling a static version of libudev (which may be a bit of work).  
3. Frankencontainer with modified rpath/LD_LIBRARY_PATH and a bunch of copied dependencies?

Or there may be another simpler way to solve this.

@smlx  - do you have any ideas on this one?  Is `libudev` strictly required if using `hidraw`?

smlx:
Yes I believe so. You'll probably need to switch to an `alpine` image for Linux.

mapkon:
> Yes I believe so. You'll probably need to switch to an `alpine` image for Linux.

I wouldn't worry about switching to alpine images. We need something light-weight, not prohibitively light weight. @gliptak - what do you think?

gliptak:
I recommend separate PR for the `goreleaser` bump

Submitted #1023 which can be used to build on

smlx:
Well in case it helps I got an alpine based image working on Linux with this Dockerfile:

```dockerfile
FROM golang:1-alpine3.17 AS builder

RUN cd / && \
      wget https://github.com/goreleaser/goreleaser/releases/download/v1.16.2/goreleaser_Linux_x86_64.tar.gz \
      && tar -xzf goreleaser_Linux_x86_64.tar.gz

RUN apk add --no-cache \
      alpine-sdk \
      linux-headers \
      eudev-dev

COPY . /saml2aws

RUN cd /saml2aws \
      && ln -fs /goreleaser ./bin/goreleaser \
      && make build

FROM alpine:3.17

RUN apk add --no-cache \
      eudev

COPY --from=builder \
      /saml2aws/dist/saml2aws_linux_amd64_v1/saml2aws \
      /saml2aws

ENTRYPOINT ["/saml2aws"]
```

Run it like this:

```
docker run --rm -it --device /dev/hidraw7 -v $HOME/.saml2aws:/root/.saml2aws -v $HOME/.aws:/root/.aws saml2aws login --disable-keychain -a exampleAccount
```

This image weighs 24.7MB vs the distroless image which is 41.1MB (when built on my machine).

BenPhegan:
Ok, a few thoughts after looking at this:

1. Neither suggested Dockerfile will work in their current form.  Both expect the source code to be available as part of the Docker build context.  However, with `goreleaser` you only get a binary from the `builds:` that matches your `goarch/goos`.  There is the option of adding `extra_files` to the docker build, however this does not accept globs.  So, using default `goreleaser` config it does not seem a simple process to _build_ the executable within docker.
2. `goreleaser` does not allow an arbitrary build process to be run as a build.  If we could, we could just build and drop out the MUSL/Alpine build and then just pick it up later.
3. We could build the Dockerfile (any arbitrary one really) from within the Makefile, and drop out a build target file as part of the Makefile target.  However this would create a fragile contract between `goreleaser` and the Makefile, based on a location in the `dist` directory and a bunch of arbitrary `goos/goarch` values.

Suggested ways forward:
1. We add an _additional_ set of build definitions to the existing `goreleaser` definition that build a _static_ version of an `amd64` binary in addition, and just forego the added capabilities (also clearly stating this in documentation).  This would mean that there would be two builds for linux/amd64, one static and one dynamic.  We then use the static binary for the Docker builds.  I think this is probably the easiest way?
2. Package in a minimal ubuntu image.  That way, nothing changes on the build or the application functionality, but you get a big container download.
3. Add in a bunch of Makefile special sauce and `goreleaser` tweaks to try and fool the Docker build process into packaging a binary build externally to `goreleaser` via the `extra_files` functionality.
4. Completely decouple the Docker build from `goreleaser` and just use a Makefile target.

I will look to implement 1. and see how I go.  I have most of this working locally at the moment, with a single `goreleaser` definition building both static and dynamic, so it seems that it should slot in nicely and _also_ provide people with a statically compiled binary option as well for those that do not need the newly added functionality that requires CGO.

Let me know if you guys have strong opinions either way though.

mapkon:
> Ok, a few thoughts after looking at this:
> 
> 1. Neither suggested Dockerfile will work in their current form.  Both expect the source code to be available as part of the Docker build context.  However, with `goreleaser` you only get a binary from the `builds:` that matches your `goarch/goos`.  There is the option of adding `extra_files` to the docker build, however this does not accept globs.  So, using default `goreleaser` config it does not seem a simple process to _build_ the executable within docker.
> 2. `goreleaser` does not allow an arbitrary build process to be run as a build.  If we could, we could just build and drop out the MUSL/Alpine build and then just pick it up later.
> 3. We could build the Dockerfile (any arbitrary one really) from within the Makefile, and drop out a build target file as part of the Makefile target.  However this would create a fragile contract between `goreleaser` and the Makefile, based on a location in the `dist` directory and a bunch of arbitrary `goos/goarch` values.
> 
> Suggested ways forward:
> 
> 1. We add an _additional_ set of build definitions to the existing `goreleaser` definition that build a _static_ version of an `amd64` binary in addition, and just forego the added capabilities (also clearly stating this in documentation).  This would mean that there would be two builds for linux/amd64, one static and one dynamic.  We then use the static binary for the Docker builds.  I think this is probably the easiest way?
> 2. Package in a minimal ubuntu image.  That way, nothing changes on the build or the application functionality, but you get a big container download.
> 3. Add in a bunch of Makefile special sauce and `goreleaser` tweaks to try and fool the Docker build process into packaging a binary build externally to `goreleaser` via the `extra_files` functionality.
> 4. Completely decouple the Docker build from `goreleaser` and just use a Makefile target.
> 
> I will look to implement 1. and see how I go. I have most of this working locally at the moment, with a single `goreleaser` definition building both static and dynamic, so it seems that it should slot in nicely and _also_ provide people with a statically compiled binary option as well for those that do not need the newly added functionality that requires CGO.
> 
> Let me know if you guys have strong opinions either way though.

@BenPhegan 

I am not sure what the appetite for this is, but I don't suppose that having a big container download is a big let off. I would seriously consider option 2 above, since it is the least intrusive from a maintenance perspective.

BenPhegan:
I actually went with static as a first pass, as I noticed there was some appetite for this outside of the use of Docker.  Let me know if this PR makes sense now, happy to discuss further and adjust as required.

BenPhegan:
Ok, rebased due to conflicts in master.  It seems that reproducibility is drifting as the Github Actions are defining versions that are not reproducible locally, including Golang and Goreleaser versions.  Is there an intent to bring this back in line and share across build mechanisms?

mapkon (commented):
@BenPhegan - I am a little confused. What is the relationship between this and #1023?

BenPhegan:
@mapkon - put a comment on the other PR, but they take different approaches.  I would suggest closing one or the other, and merging the efforts.  My preference is an approach that includes goreleaser compatibility, but that is only based on the fact that it appears to be the primary mechanism the project is using to build/deploy currently.

BenPhegan:
@mapkon - put a comment on the other PR, but they take different approaches.  I would suggest closing one or the other, and merging the efforts.  My preference is an approach that includes goreleaser compatibility, but that is only based on the fact that it appears to be the primary mechanism the project is using to build/deploy currently.

BenPhegan:
@mapkon - outstanding issues with this PR:

1. Acceptance that this change creates both _dynamic_ and _static_ binaries for Linux, where the static linux binaries have reduced capabilities.
2. Documentation should be updated to reflect this, if we accept 1. above.  Happy to update as necessary if you are ok with the static binary approach.
3. `goreleaser` Docker manifest creation and publishing have _not_ been tested, as they require a Docker Hub account registered to Versent and a full deployment to be run.  There may be tweaks associated with this that I just cannot test.
4. Optimisation would be to remove the `.goreleaser.macos-latest.yml` definition completely and just use a single definition file/build.  I can do this as a seperate PR as well.

As this is being used in preference to a single Dockerfile as per the linked PR, it would be good to get feedback from @gliptak to make sure he is comfortable with this approach as well?

gliptak:
maybe merge #1010 and start working Dockerfile/etc. after?

@BenPhegan we might also consider publishing to ghcr.io

https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L83

mapkon:
> @mapkon - outstanding issues with this PR:
> 
> 1. Acceptance that this change creates both _dynamic_ and _static_ binaries for Linux, where the static linux binaries have reduced capabilities.
> 2. Documentation should be updated to reflect this, if we accept 1. above.  Happy to update as necessary if you are ok with the static binary approach.
> 3. `goreleaser` Docker manifest creation and publishing have _not_ been tested, as they require a Docker Hub account registered to Versent and a full deployment to be run.  There may be tweaks associated with this that I just cannot test.
> 4. Optimisation would be to remove the `.goreleaser.macos-latest.yml` definition completely and just use a single definition file/build.  I can do this as a seperate PR as well.
> 
> As this is being used in preference to a single Dockerfile as per the linked PR, it would be good to get feedback from @gliptak to make sure he is comfortable with this approach as well?

The management of docker credentials really worries me. Is there a way we can avoid that?

mapkon:
> maybe merge #1010 and start working Dockerfile/etc. after?
> 
> @BenPhegan we might also consider publishing to ghcr.io
> 
> https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L83

There is no PR on #1010

gliptak:
@mapkon ghcr.io requires no additional credentials

BenPhegan:
@mapkon - ok, I will hold off as I will need to redo the PR based on top of #1029 (which actually ends up in a similar place as this one I believe).  There will be conflicts due to the duplication.  I will pick this up again when #1029 is complete.  Also, yes it makes complete sense to use ghcr.io, I will update PR.

mapkon:
@BenPhegan #1029 is merged, and I have closed #1023 so we can focus on this.

BenPhegan:
Ok, great, I will try and get some time to tidy this (rebase etc) today.

BenPhegan:
Took me a bit longer than expected to get to this, but I _think_ this is looking reasonable now, with the caveats that the Docker manifest publishing has _not_ been tested...

mapkon:
> Took me a bit longer than expected to get to this, but I _think_ this is looking reasonable now, with the caveats that the Docker manifest publishing has _not_ been tested...

Thank you. What is remaining here for us to be ready to merge?

BenPhegan:
@mapkon the release process remains to be tested, which I cannot do as I do not have the ability to generate a token for the project to push to ghcr.io.  There are probably some tweaks to the documentation depending on that release process as well.  So two approaches we could take:

1. Someone with the ability to generate a release environment similar to the Github Actions environment for the branch pulls this locally, sets up the environment and tests the release process from their machine, and adjusts anything that does not work accordingly.
2. We merge it, and I follow up with a PR fixing any release based issues if they arise.

I think 2. is probably easier, if you are happy to give that a shot.  If you let me know when it is going to be merged I can try to be on line to ensure I can respond quickly to any issues with the docker manifest deployment process.

gliptak:
ghcr.io relies on `secrets.GITHUB_TOKEN`

you might consider running (ghcr.io) release on your fork

mapkon:
> ghcr.io relies on `secrets.GITHUB_TOKEN`
> 
> @BenPhegan consider running (ghcr.io) release on your fork

@BenPhegan Please see @gliptak reply. We should test on your branch/fork before merging to master. That way, we can limit the blast radius.

BenPhegan:
@gliptak - if I run on my fork, it will still be attempting to public an image to ghcr.io/versent/saml2aws.  Do you want me to abstract that publish location to accept forks?  Perhaps I am missing something in how the publish could work from my fork?

gliptak:
@BenPhegan maybe I'm wrong and package name is "hardcoded" for golang/goreleaser

mapkon:
@BenPhegan / @gliptak  - is this ready to be merged?

gliptak:
@BenPhegan did you succeed publishing at your own fork?

BenPhegan:
@mapkon / @gliptak - no, I have not been able to publish to my own fork.  This has become a massive time sink. I have been blocked by an error as you can see here: https://github.com/BenPhegan/saml2aws/actions/runs/4932162770/jobs/8814923274

If I look at the documentation here https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/:

> The GITHUB_TOKEN is an automatically generated secret that lets you make authenticated calls to the GitHub API in your workflow runs. Actions generates a new token for each job and expires the token when a job completes. The token has write permissions to a number of [API endpoints](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) except in the case of pull requests from forks which are always read. These new settings allow you to follow a principle of least privilege in your workflows.

To pull out the important bit, **except in the case of pull requests from forks which are always read**. So basically I do not think it is possible to do full testing on a fork as a fork will only ever have a read access token.  I can remove the fork and republish, but that kinda removes the point.  Unless I am missing something here (quite possible, please let me know if you have done this before and can let me know how).

Up to you guys really regarding ready to merge.  It seems to work up until the point of trying to publish.  That code has not been tested against another repo ownership (it is attempting to use the name of the repository owner as part of the docker image name rather than static, as requested).  At this point, it is really just a case of either closing this one, or merging it and fixing any issues that come up.  I am more than happy to help with this, but I don't think I can justify spending any more time on the PR itself.

gliptak:
@BenPhegan thank you for working through this

#1053 was based on your work just with minimal changes, please rebase against

somehow MSI generation was impacted. maybe another fork issue?

gliptak:
@BenPhegan yes this PR can be closed. Next release/tag expected to publish ghcr.io/versent/saml2aws. Thank you again for working this

mapkon:
@BenPhegan If you want a release, please huddle a testing run - here is an [example](https://github.com/Versent/saml2aws/discussions/1038)

BenPhegan:
@mapkon - sure, happy to get involved.  However, there is a chicken/egg problem here. As the docker images are actually part of the _release_ process and goreleaser does not create them until you _publish_, we can test everything but the Docker images as part of a pre-release process.  Not sure if there is a way around this.  I can pull down master and test the locally built Docker images to ensure that they work though.

gliptak:
@BenPhegan I just published HEAD as v2.37.6

https://github.com/gliptak/saml2aws/pkgs/container/saml2aws
https://github.com/gliptak/saml2aws/actions/runs/5086259720

https://github.com/gliptak/saml2aws/actions/runs/5086251019

Feel free to use it for testing huddle

BenPhegan:
Raised discussion here: https://github.com/Versent/saml2aws/discussions/1065

Closing this PR.

PR #780, opened 2022-02-25
onelogin: use the /api/2/saml_assertion API

rodmatos:
Replaces the use of the /api/1/saml_assertion with its successor.
OneLogin has communicated that the /api/1/saml_assertion has been
deprecated and should have been removed in Feb 2021.
Fixes issue #778.

[Documentation for the /1/saml_assertion](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion).
[Documentation for the new /2/saml_assertion](https://developers.onelogin.com/api-docs/2/saml-assertions/generate-saml-assertion).

PR #904, opened 2022-11-01
Ptsier/onelogin saml v2

### LINKED ISSUE #778: OneLogin API /1 is deprecated
I see that saml2aws is using the `/1` version of OneLogin's API:

https://github.com/Versent/saml2aws/blob/cdd6da80f128dd17d400251b08ea5d226bdfe154/pkg/provider/onelogin/onelogin.go#L116

[The API docs](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion) indicate that this endpoint shouldn't be used:

> Deprecated
> This version is deprecated and will be removed in February 2021.

I've looked at [the `/2` api docs](https://developers.onelogin.com/api-docs/2/saml-assertions/generate-saml-assertion) and the responses seem to be structured a bit differently.

I haven't reached out to OneLogin yet about if/when they plan to actually remove the `/1` API.

Barraketh:
Fixes #778 .  @rodmatos has an open PR to fix this which this PR is based on (https://github.com/Versent/saml2aws/pull/780), but that PR is slightly incorrect (see commit 2fd27b1835d51ce3847e2fae3aa094592a7f09eb).

The v1 api is still available, but has started to break randomly at our company, so updating to v2 seems worthwhile.

[Documentation for the /1/saml_assertion](https://developers.onelogin.com/api-docs/1/saml-assertions/generate-saml-assertion).
[Documentation for the new /2/saml_assertion](https://developers.onelogin.com/api-docs/2/saml-assertions/generate-saml-assertion).

### DIFF
--- a/pkg/provider/onelogin/onelogin.go
+++ b/pkg/provider/onelogin/onelogin.go
@@ -112,7 +112,7 @@ func (c *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) 
 		return "", errors.Wrap(err, "error encoding authreq")
 	}
 
-	authSubmitURL := fmt.Sprintf("https://%s/api/2/saml_assertion", host)
+	authSubmitURL := "https://api.us.onelogin.com/api/2/saml_assertion"
 
 	req, err := http.NewRequest("POST", authSubmitURL, &authBody)
 	if err != nil {


rodmatos:
@Barraketh: I am not sure if I fully understand the problem that you have identified on the pull request (#780) that I have raised. Could you elaborate a bit more please?

rodmatos:
Still not sure if I understand, isn't subdomain here `api.us.onelogin.com|api.eu.onelogin.com`?

Barraketh:
You're right.  I'll close this PR, but just to explain my confusion:

So in the OneLogin documentation, the subdomain is unique to your organization (see https://developers.onelogin.com/api-docs/2/getting-started/dev-overview ) so it would be something like "myorg.onelogin.com" .

We are making two calls: one to `[url]/auth/oauth2/v2/token` to get the token, and one to `[url]/api/2/saml_assertion` .  When url is set to `myorg.onelogin.com`, the first call succeeds but the second fails. So I figured that we need two different urls - one for the first call which would be `myorg.onelogin.com`, and one for the second which would be `api.us.onelogin.com`.  However, turns out that setting the url to `api.us.onelogin.com` works for both calls, so we don't need to differentiate.

PR #907, opened 2022-11-03
Replace browser URL listener with Regex to support GovCloud

### LINKED ISSUE #905: [BUG] Browser login does not support GovCloud
**ISSUE:** Currently the browser login waits for a signin request to the URL: `https://signin.aws.amazon.com` but when signing into AWS GovCloud the signin URL is: `https://signin.amazonaws-us-gov).com` so the signin is never detected.

**SUGGESTION:** The string used to listen for the proper URL could be replaced with a regex that could support both the standard signin URL and the GovCloud URL.

zacklmoore:
Fixes #905


mapkon:
@zacklmoore The linter is complaining about the use of strings in the code. Would you kindly take a look?

gliptak:
this seem to be covered at https://github.com/Versent/saml2aws/blob/master/pkg/provider/browser/browser.go#L99

mapkon:
Right! Closing as duplicate of #992

PR #934, opened 2023-01-16
Add Authentik provider

ttitoo:
[Authentik](https://goauthentik.io/) support, without MFA support, compatible with 2022.12.0(maybe old versions, not sure), up to 2022.12.2.

### DIFF
--- a/pkg/provider/authentik/authentik.go
+++ b/pkg/provider/authentik/authentik.go
@@ -1,0 +1,284 @@
+package authentik
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/versent/saml2aws/v2/pkg/cfg"
+	"github.com/versent/saml2aws/v2/pkg/creds"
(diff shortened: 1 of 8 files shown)


mapkon:
@ttitoo There are conflicts with master. Can you take a look? I would like to merge this ASAP

ttitoo:
I'll fix the error today

ttitoo:
I think it's working @mapkon

ttitoo:
I think it's working now @mapkon

pkg/provider/authentik/authentik.go
```diff
@@ -0,0 +1,284 @@
...
+	b, err := ioutil.ReadAll(res.Body)
+	err = json.Unmarshal(b, &payload)
+	if err != nil {
+		return nil, err
+	}
+
+	return &payload, nil
+}
+
+func aaa(a int) int {
```
mapkon, line 282: Is this intentional? Could we find a better function name?

mapkon:
```
  Error: `aaa` is unused (deadcode)
  Error: File is not `goimports`-ed (goimports)
  Error: ineffectual assignment to err (ineffassign)
  Error: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now
```

ttitoo: It's a mistake, removed

PR #951, opened 2023-02-09
Support U2F on Linux

smlx:
This PR enables U2F support on Linux amd64. It builds on the work done in, and supersedes, #592.

There are no code changes, just build configuration updates:

* Pass `-tags=hidraw` to `go build` on Linux amd64, which enables U2F as per [the documentation](https://github.com/marshallbrekka/go-u2fhost#linux) in the underlying U2F library.
* Install the `libudev-dev` package and use `CGO_ENABLED=1` on Linux, because U2F support uses libudev.
* Update the Github workflows to build the package correctly on Linux (untested).
* Update the build documentation in the README.

Closes: #419 
Closes: #592

Note that this changes the release builder from macos to ubuntu. I have no way of testing the resulting macos binary to see if it still works, so would appreciate if someone else could confirm this.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -61,6 +61,9 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Install dependency required for linux builds
+      run: sudo apt-get update && sudo apt-get install -y libudev-dev
+
     - name: GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,12 +3,12 @@ name: release
 on:
   push:
     tags:
-      - '*'
+    - '*'
 
 jobs:
   release:
     name: release
-    runs-on: macOS-latest
+    runs-on: ubuntu-latest
     steps:
 
     - name: Set up Go 1.x
@@ -18,6 +18,9 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+
+    - name: Install dependency required for linux builds
+      run: sudo apt-get update && sudo apt-get install -y libudev-dev
 
     - name: GoReleaser
       uses: goreleaser/goreleaser-action@v2
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -17,6 +17,14 @@ builds:
     - amd64
     - arm64
     - arm
+  overrides:
+    - goos: linux
+      goarch: amd64
+      goamd64: v1
+      tags:
+        - hidraw
+      env:
+        - CGO_ENABLED=1
 archives:
   - format: tar.gz
     wrap_in_directory: false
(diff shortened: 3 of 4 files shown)


rarruda:
@mapkon @wolfeidau any chances this PR could get reviewed/merged? 

It is an important addition to us that run on Linux. Also, this PR seems pretty clean: only has changes to get U2F working on linux, and nothing more.

.github/workflows/go.yml
```diff
@@ -61,6 +61,9 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Install dependency required for linux builds
+      run: sudo apt-get update && sudo apt-get install -y libudev-dev
```
mapkon, line 65: Why do you need `libudev-dev`?

.github/workflows/release.yml
```diff
@@ -19,6 +19,9 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Install dependency required for linux builds
+      run: sudo apt-get update && sudo apt-get install -y libudev-dev
```
mapkon, line 23: Ditto here

mapkon (approved):

.github/workflows/go.yml
rarruda: because U2F support uses libudev.

mapkon:
I think we should move all the build configuration files to run on `ubuntu-latest`. `MacOS` builders have a 10x multiplier: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers

.github/workflows/go.yml
mapkon: OK! The branch is a bit out of date with master. Mind rebasing?

rarruda:
Makes sense, but I'd say that's outside the scope of this PR. :slightly_smiling_face:

mapkon:
> Makes sense, but I'd say that's outside the scope of this PR. 🙂

I don't think it will merge! I will try to get it merged ASAP!

rarruda:
I didn't create the PR, so I don't think I can rebase in this PR. But I could create another PR (from another frok) being a rebase of this one. Would that work?

Or would you prefer to try "rebase and merge" from github or doing it directly yourself?

Just let me know how I can help!

.github/workflows/go.yml
rarruda: see https://github.com/Versent/saml2aws/pull/951#issuecomment-1447793461 :slightly_smiling_face:

smlx:
I can rebase it :slightly_smiling_face:

rarruda:
Thanks @smlx !

mapkon:
@smlx @rarruda 

Please note that using `ubuntu` platforms to build the release breaks OS builds. I am reverting that for now. @wolfeidau and I will take another peek to fix that properly, for now, we reverting back to `macos-latest` to lessen the blast radius.

smlx:
@mapkon Ah I was afraid of that, I'll send another PR to fix it up.

mapkon:
Not a worry! Let me know when the PR is ready.

PR #961, opened 2023-02-28
 Revert the release build to use macos-latest

mapkon:
- This functionality was broken in #951
 - Since we are using OSX toolchain, we need macos-X to build


wolfeidau (approved):
👍🏻 🚀

PR #966, opened 2023-03-04
Bump Golang to 1.20

gliptak:


mapkon:
@gliptak This will need to be updated due to the merging of the PR that bumps Golang to 1.9X

Thank you a bunch for your contributions.

PR #988, opened 2023-03-16
Add Codecov to GHA

gliptak:
might have to be merged (so primary branch baseline is uploaded) before fully functional (or might require followup tweaks)

https://app.codecov.io/github/Versent/saml2aws/pull/988


mapkon (approved):
Thanks mate! This is important especially given that we are pushing for better tests and code coverage

PR #991, opened 2023-03-17
Ignore (generated) coverage.xml

gliptak:
Ignore (generated) coverage.xml


mapkon:
Mind explaining why we are ignoring `coverage.xml`?

PR #957, opened 2023-02-21
Okta MFA - Support multiple of same type

szechyjs:
This resolves the linting issues in #745 since its has gone stale (abandoned?).

PR #998, opened 2023-03-22
split coverage report generation and upload

briantist:
Demonstration of what I described here:
- https://github.com/Versent/saml2aws/discussions/996#discussioncomment-5387761


.github/workflows/go.yml
```diff
@@ -28,13 +28,13 @@ jobs:
...
 
-    - name: Codecov
-      uses: codecov/codecov-action@v3
+    - name: Upload coverage report
+      uses: actions/upload-artifact@v3
       with:
-        file: coverage.txt
-        flags: unittests
+        name: reports
+        path: ${{ matrix.os }}_coverage.txt
```
briantist, line 37: use short retention and ensure no coverage file is an error

```suggestion
        name: reports
        path: ${{ matrix.os }}_coverage.txt
        if-no-files-found: error
        retention-days: 1        
```

briantist:
While it all appears to be successful, the two runs so far had errors processing on the codecov side for some reason:
- [[run](https://github.com/Versent/saml2aws/actions/runs/4485824731)] -- [[codecov](https://app.codecov.io/github/Versent/saml2aws/commit/f0ea15ed8254198677449c17472f6d36e6bcd1b0)]
- [[run](https://github.com/Versent/saml2aws/actions/runs/4485854428)] -- [[codecov](https://app.codecov.io/github/Versent/saml2aws/commit/810dfaf440dc26dd1abb63ec340e59b341b28c21)]

I'm not sure why that is; it might be that they actually do rely on `contents: read` but don't document it (but https://github.com/codecov/codecov-action/issues/739 suggests otherwise).

It may also be that the location of the coverage file matters, and while it was in the root before, I'm downloading it into a directory.

I did confirm the uploaded reports have content, and they do seem to be uploaded to codecov according to the action's output.

.github/workflows/go.yml
```diff
@@ -57,6 +59,25 @@ jobs:
       with:
         version: v1.51.2
 
+  coverage:
+    name: coverage
+    permissions:
```
briantist, line 64:
```suggestion
    permissions:
      contents: read
```

.github/workflows/go.yml
```diff
@@ -57,6 +59,26 @@ jobs:
...
+      contents: read
+      actions: read
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
```
briantist, line 69:
```suggestion
    steps:
    - uses: actions/checkout@v3
```

briantist:
looking at the content of the uploaded reports (to codecov, not to the workflow artifacts) from successful (other runs) vs. this, I suspect codecov requires the repo to be checked out, so even though their uploader doesn't require `contents: read`, we do need that permission to perform the checkout.

Let's see if that does the trick in https://github.com/Versent/saml2aws/pull/998/commits/8cfef607e1f915d261e760e49837895f51a04071

mapkon (approved):

briantist:
@mapkon setting this back to draft because although it appears successful, the last two runs did not update the codecov comment, and when I look at them, they don't show an error but appear to be still processing, very strange:
- https://app.codecov.io/github/Versent/saml2aws/commit/db7273658c97c5f135643331559093be3611e910
- https://app.codecov.io/github/Versent/saml2aws/commit/5a8b9947a340ffd6e421c60510b76d330215d766

briantist:
hm, this might be a problem on codecov's end... for example the latest run in #957 has the same issue:
- https://app.codecov.io/github/Versent/saml2aws/commit/648be5e611117ac5869bb3600d10370a4a171cea

(and if you look at the edit time on the codecov comment in that PR, it too has not been updated)

mapkon:
Yeah, took a punt at merging but the reports look iffy, then I demurred. I am glad you noticed the strangeness too -

.github/workflows/go.yml
```diff
@@ -28,13 +28,15 @@ jobs:
       run: |
         go install github.com/playwright-community/playwright-go/cmd/playwright
         playwright install --with-deps
-        go test -v ./... -coverprofile=coverage.txt -covermode=atomic
+        go test -v ./... -coverprofile=${{ matrix.os }}_coverage.txt -covermode=atomic
 
-    - name: Codecov
-      uses: codecov/codecov-action@v3
+    - name: Upload coverage report
```
gliptak, line 33: @briantist please add below after this line

```
if: matrix.os == 'ubuntu-18.04'
```

.github/workflows/go.yml
jsoref: Fwiw, this does a couple of things and should be included in `build:` as well...

specifically, if you don't specify any permissions, each workflow/job will default to some level of maximal permissions depending on various things. By specifying _any_ permissions, the permissions categories you don't specify get no permission (metadata being a caveat, but...) -- this is a good thing -- you're limiting permissions to only those you ask for, instead of unintentionally allowing potentially quite generous permissions.

.github/workflows/go.yml
```diff
@@ -57,6 +59,26 @@ jobs:
...
+        path: reports
+
+    - name: Codecov
+      uses: codecov/codecov-action@v3
+      with:
+        directory: reports
+        flags: unittests
+
   release-build:
     name: release-build
```
jsoref, line 83: This job should have:
```
permissions:
  contents: write
```
per https://github.com/goreleaser/goreleaser-action#workflow

briantist: I didn't change the permissions for that job in this PR; when it runs via `push` trigger it will have write access in all the scopes. If we want to pare down permissions for this job to only `contents: write` (it's a good idea!) we should do it in a different PR.

briantist: one other thing is that we can't actually give it `contents: write` for the `pull_request`-triggered runs if the PR is based on a fork, so the job will actually fail in that case... I don't know what goreleaser is actually doing during PR runs, I guess it's just ensuring it can build and such since it doesn't do an actual release, and in that case it won't need write.

briantist:
> Yeah, took a punt at merging but the reports look iffy, then I demurred. I am glad you noticed the strangeness too -

It looks like it was indeed a problem on codecov's end; both this PR and others in this repo are working fine now, so I think we're good with this!

This latest lint failure does not appear to be related to this PR (also happening in others, see https://github.com/Versent/saml2aws/actions/runs/4499509224)

.github/workflows/go.yml
gliptak: @briantist thank you. I proposed as it corrected Codecov reporting in my setup. Do reports have unique names (based on OS)?

mapkon: I suppose based on how they are being uploaded, yes! 
`go test -v ./... -coverprofile=${{ matrix.os }}_coverage.txt -covermode=atomic`

gliptak: @mapkon @briantist I didn't catch that update. thx

PR #1001, opened 2023-03-23
[AzureAD] Fix broken HTML parser of AzureAD integration

kstkn:
JSON that is embedded in a script tag can have ";" inside which breaks the parser. This change doesn't make the parser cleaner but at least handles such cases properly

### DIFF
--- a/pkg/provider/aad/aad.go
+++ b/pkg/provider/aad/aad.go
@@ -15,6 +15,7 @@ import (
 	"github.com/PuerkitoBio/goquery"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+
 	"github.com/versent/saml2aws/v2/pkg/cfg"
 	"github.com/versent/saml2aws/v2/pkg/creds"
 	"github.com/versent/saml2aws/v2/pkg/prompter"
@@ -670,8 +671,9 @@ func (ac *Client) getJsonFromConfig(resBodyStr string) string {
 	 * <script><![CDATA[  $Config=......; ]]>
 	 */
 	startIndex := strings.Index(resBodyStr, "$Config=") + 8
-	endIndex := startIndex + strings.Index(resBodyStr[startIndex:], ";")
-	return resBodyStr[startIndex:endIndex]
+	endIndex := startIndex + strings.Index(resBodyStr[startIndex:], "//]]>") - 1
+	untrimmedJsonStr := resBodyStr[startIndex:endIndex]
+	return untrimmedJsonStr[:strings.LastIndex(untrimmedJsonStr, "}")+1]
 }
 
 func (ac *Client) responseBodyAsString(body io.ReadCloser) (string, error) {
--- a/pkg/provider/aad/testdata/LoginEmbeddedJsonLineBreak.html
+++ b/pkg/provider/aad/testdata/LoginEmbeddedJsonLineBreak.html
@@ -1,0 +1,11 @@
+  <meta name="robots" content="none" />
+
+  <script type="text/javascript">//<![CDATA[
+  $Config={"name": "value&amp;with ampersand"};//]]></script>
+  <script type="text/javascript">//<![CDATA[
+
+  //]]>
+  </script>
+
+</body>
+</html>
--- a/pkg/provider/aad/aad_test.go
+++ b/pkg/provider/aad/aad_test.go
@@ -7,10 +7,12 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"math"
 	mrand "math/rand"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"strconv"
 	"sync"
 	"testing"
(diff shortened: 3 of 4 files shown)


kstkn:
@christianmeyer thanks for seeing this. 
This fix is definitely not a long term solution but it at least resolves this edge case where for example some companies with names like "Something & something" are struggling right now :roll_eyes:

pkg/provider/aad/testdata/LoginEmbeddedJsonLineBreak.html
puellanivis, line 1: The names of these two test files seem to be backwards?

pkg/provider/aad/aad.go
```diff
@@ -670,8 +671,9 @@ func (ac *Client) getJsonFromConfig(resBodyStr string) string {
 	 * <script><![CDATA[  $Config=......; ]]>
 	 */
 	startIndex := strings.Index(resBodyStr, "$Config=") + 8
-	endIndex := startIndex + strings.Index(resBodyStr[startIndex:], ";")
-	return resBodyStr[startIndex:endIndex]
+	endIndex := startIndex + strings.Index(resBodyStr[startIndex:], "//]]>") - 1
+	untrimmedJsonStr := resBodyStr[startIndex:endIndex]
+	return untrimmedJsonStr[:strings.LastIndex(untrimmedJsonStr, "}")+1]
```
puellanivis, line 676: Why don’t we take advantage of the fact that a streaming JSON parser will return the first parsed JSON entity and ignore any additional data following it:
```
var msg json.RawMessage
_ = json.NewDecoder(strings.NewReader(resBodyStr[startIndex:])).Decode(&msg)
return string(msg)
```

Otherwise, the code as proposed could break if there are additional lines of javascript, e.g. `$Config={...}; if (false) {  alert("oh noes!") };` Best of all, _no more ad hoc self parsing_.

pkg/provider/aad/testdata/LoginEmbeddedJsonExtraJavascript.html
```diff
@@ -0,0 +1,12 @@
+  <meta name="robots" content="none" />
+
+  <script type="text/javascript">//<![CDATA[
+  $Config={"name": "value&amp;with ampersand"}; alert('oops');
```
puellanivis, line 4: Since your original design used `strings.IndexLast(config, "}")` it shouldn’t have broken on this. This is why I had the `if (false) { … }`, I had to have a `}` after the end of the JSON.

kstkn: true, this test is useless

christianmeyer:
@kstkn I was working on a similar approach and we both ended up with similar outcomes, I tried to eliminate the String position guess work overall ... but your solution looks simpler .. as long as the JsonParser can handle the overhead as you pass the overall HTML page from the start position.

Here's my version merged with your function refactoring (not yet fully tested) -- pick if you like
```
func (ac *Client) unmarshalEmbeddedJson(resBodyStr string, v any) error {
    /*
     * data is embedded in a javascript object
     * <script><![CDATA[  $Config=......; ]]>
     */
    var html *goquery.Document
    var err error
    html, err = goquery.NewDocumentFromReader(strings.NewReader(resBodyStr))
    if err != nil {
        return errors.Wrap(err, "error parsing HTML body for $Config JSON")
    }
    configJson := func(i int, s *goquery.Selection) bool {
        return strings.Contains(s.Text(), "$Config=")
    }
    configString := html.Find("script").FilterFunction(configJson).First().Text()
    if len(strings.TrimSpace(configString)) > 0 {
        configString = strings.TrimPrefix(strings.TrimSpace(configString), "//<![CDATA[")
        configString = strings.TrimSuffix(strings.TrimSpace(configString), "//]]>")
        configString = strings.TrimPrefix(strings.TrimSpace(configString), "$Config=")
        configString = strings.TrimSuffix(strings.TrimSpace(configString), ";")
        return json.NewDecoder(strings.NewReader(strings.TrimSpace(configString))).Decode(&v)
    } else {
        return fmt.Errorf("unable to decode $Config JSON")
    }
}
```

kstkn:
I personally think the outweighing advantage of `json.NewDecoder` over `Unmarshal` is that it will decode first JSON-encoded value which makes this otherwise crufty solution somewhat more robust. And having in mind that this eventually can break again, leaving this as simple as possible will be the most optimal.

mapkon:
> @christianmeyer now we actually made it better (even though we're still extracting a JSON from HTML page 😬 ) @mapkon could you please consider merging this and releasing a patch version

Sure thing buddy! I will take a look and circle the wagons

danishnawab:
@mapkon when do you plan to release this?

mapkon:
> @mapkon when do you plan to release this?

We are organizing a test of the current codebase. Please see here: https://github.com/Versent/saml2aws/pull/1029#issuecomment-1533924283

If you can spend sometime to test out the current master and ascertain release readiness, I will appreciate

PR #992, opened 2023-03-18
Add browser test

gliptak:
#937 

Had to switch to `ubuntu-18.04` as `github.com/playwright-community/playwright-go/cmd/playwright` fails to install on `20.04+`

https://github.com/playwright-community/playwright-go/issues/330
https://github.com/playwright-community/playwright-go/issues/335

relatedly https://github.com/microsoft/?q=playwright&type=all&language=&sort= seems to have more current `playwright` repos except Golang https://github.com/microsoft/playwright/issues/6856

https://github.com/playwright-community/playwright-go

PR #1002, opened 2023-03-23
Refs #992: Use a valid saml response for tests

mapkon:


mapkon:
@gliptak Please take a look 🗡️

gliptak:
:+1:

PR #982, opened 2023-03-11
Move playwright-go from mxschmitt to playwright-community

gliptak:
#952 

I didn't test browser path of resulting binary

Issue #952, opened 2023-02-09
Browser provider downloads old version of playwright chromium


mikex99:
When using the browser provider, chromium v907428 is downloaded.  This is an out of date version.

PR #786, opened 2022-03-10
Add the option to set a persistent storage for the browser provider

gngj:
This enables using an IDP that has a browser session and providing credentials only once.

Can help #729

Example usage:
`saml2aws configure --url https://account.activedirectory.windowsazure.com/applications/signin/<app_id>?tenantId=<tenantId> --persistent-data-dir ~/.persistent_data_dir -a browser --idp-provider Browser --skip-prompt`

PR #1006, opened 2023-03-26
Browser IDP - Configure Browser IDP to automatically install Playwright Browser Drivers per old behavior <= 2.36.4

### LINKED ISSUE #1005: Getting playwright error when trying to saml2aws login
Hey,

Getting an error when I login to saml2aws.

Error: 
```
~ saml2aws login
Using IdP Account default to access Browser `URL`
To use saved password just hit enter.
? Username `username`
? Password

Authenticating as `username` ...
Error authenticating to IdP.: could not start driver: fork/exec /Users/`username`/Library/Caches/ms-playwright-go/1.20.0-beta-1647057403000/playwright.sh: no such file or directory
```

Steps:
```
brew install saml2aws
saml2aws configure
saml2aws login
```

Version: 
```
~ saml2aws --version
2.36.5
```

xinkecf35:
# Description
This PR seeks to have the browser IDP option to once again automatically download Browser Drivers as was the default behavior for saml2aws versions <= 2.36.4

This PR would functionally close #1005 

## Motivation

When saml2aws was upgraded to 2.36.5 and using the IDP option of `Browser`, one would get this failure message:
```
Error to authenticating to IdP: could not start driver: fork/exec /Users/alice/Library/Caches/ms-playwright-go/1.20.0-beta-1647057403000/playwright.sh: no such file or directory
```

It appears that as part of upgrading browser binaries and thus playwright as part of issue #952 and PR #982 introduced a version of [playwright-go](https://github.com/playwright-community/playwright-go/releases/tag/v0.2000.0) that had breaking change to the underlying playwright driver that will no longer automatically acquire browser drivers.

Per the breaking change [notice](https://github.com/playwright-community/playwright-go/releases/tag/v0.2000.0) in playwright-go, this PR seeks to add the API Install method of acquiring the browser driver.

## Alternatives
Alternatively, instead of baking in the install, the README could be updated to reflect this requirement, and have the user on their own acquired the driver via either the playwright-go CLI or the upstream Playwright Typescript project.

This however does add  friction to use saml2aws, especially if one is using a package manager like Homebrew or Chocolatey, i.e. needing to on top of acquiring saml2aws via the documented means to also separately acquire Go Tools or NPM/NPX/Node to acquire playwright driver.

## Additional Comments
The author recognizes the implementation is somewhat naive in that it gives no configuration to avoid downloading the drivers which might desirable, e.g. avoid re-downloading drivers for CI/CD tests

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -16,22 +16,20 @@ var logger = logrus.WithField("provider", "browser")
 
 // Client client for browser based Identity Provider
 type Client struct {
-	Headless        bool
-	DownloadBrowser bool
+	Headless bool
 }
 
 // New create new browser based client
 func New(idpAccount *cfg.IDPAccount) (*Client, error) {
 	return &Client{
-		Headless:        idpAccount.Headless,
-		DownloadBrowser: idpAccount.DownloadBrowser,
+		Headless: idpAccount.Headless,
 	}, nil
 }
 
 func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) {
 
-	// Optionally download browser drivers if defined in config or set by flag
-	if cl.DownloadBrowser || loginDetails.DownloadBrowser {
+	// Optionally download browser drivers if specified
+	if loginDetails.DownloadBrowser {
 		err := playwright.Install()
 		if err != nil {
 			return "", err
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -30,8 +30,8 @@ func New(idpAccount *cfg.IDPAccount) (*Client, error) {
 
 func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) {
 
-	// Optionally download browser drivers if defined in config
-	if cl.DownloadBrowser {
+	// Optionally download browser drivers if defined in config or set by flag
+	if cl.DownloadBrowser || loginDetails.DownloadBrowser {
 		err := playwright.Install()
 		if err != nil {
 			return "", err
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -17,26 +17,33 @@ var logger = logrus.WithField("provider", "browser")
 // Client client for browser based Identity Provider
 type Client struct {
 	Headless bool
+	// Setup alternative directory to download playwright browsers to
+	BrowserDriverDir string
 }
 
 // New create new browser based client
 func New(idpAccount *cfg.IDPAccount) (*Client, error) {
 	return &Client{
-		Headless: idpAccount.Headless,
+		Headless:         idpAccount.Headless,
+		BrowserDriverDir: idpAccount.BrowserDriverDir,
 	}, nil
 }
 
 func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) {
+	runOptions := playwright.RunOptions{}
+	if cl.BrowserDriverDir != "" {
+		runOptions.DriverDirectory = cl.BrowserDriverDir
+	}
 
 	// Optionally download browser drivers if specified
 	if loginDetails.DownloadBrowser {
-		err := playwright.Install()
+		err := playwright.Install(&runOptions)
 		if err != nil {
 			return "", err
 		}
 	}
 
-	pw, err := playwright.Run()
+	pw, err := playwright.Run(&runOptions)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -26,6 +26,13 @@ func New(idpAccount *cfg.IDPAccount) (*Client, error) {
 
 func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) {
 
+	// Automatically download browser downloads to match behavior releases < 2.36.4
+	// TODO: provide override to not automatically download playwright browsers
+	err := playwright.Install()
+	if err != nil {
+		return "", err
+	}
+
 	pw, err := playwright.Run()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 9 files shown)


mapkon (changes requested):
@xinkecf35 

This is good, and it was by design that we do not impose the downloading of the playwright during the login process. From my analysis, it appears that the upstream project doesn't check or store info which allows it to invalidate the cache on upgrade, something that has been an issue on the upstream project: https://github.com/playwright-community/playwright-go/issues/236

I see two solutions here:
- Either way check that the binary exists and only download if it doesn't
- Document the presence of playwright as a requisite library
- Find a way to bundle both of them so they are installed together.

@briantist / @gliptak - any thoughts?

briantist:
Hey @mapkon I'd like to help where I can, but I've actually never used golang and I'm not really familiar with the code in this project. I do have a lot of experience with GitHub Actions though, so hopefully I can continue to assist in those areas!

mapkon:
> Hey @mapkon I'd like to help where I can, but I've actually never used golang and I'm not really familiar with the code in this project. I do have a lot of experience with GitHub Actions though, so hopefully I can continue to assist in those areas!

No worries! Thanks for clarifying.

gliptak:
including this with `saml2aws` sounds like a good approach (setting up `playwright.Install()` call at another control flow path)

reading through both flags and commands, both don't seem to offer provider specific overrides.

mapkon:
> including this with `saml2aws` sounds like a good approach (setting up `playwright.Install()` call at another control flow path)
> 
> reading through both flags and commands, both don't seem to offer provider specific overrides.

How about we document it? I am not really sure about downloading dependencies to enact the basic functionality of the program, which is login?

gliptak:
download outside `saml2aws` would require enduser installing `go`

gliptak:
other browser related #786 #729 #815

mapkon:
Ref: https://github.com/Versent/saml2aws/discussions/1011#discussioncomment-5609782

@wolfeidau We are starting to get users reporting issues related to the playwright dependency not downloading, something that affects the functioning of the app. How do you suggest we proceed on this?

xinkecf35:
Just to provide a suggestion forward:

Perhaps add a new configuration property to the config called `download_browser_driver` to choose to allow saml2aws to install the playwright browser for you with the new default going forward being not to?

Or maybe with a flag on `saml2aws login` called `--download-browser-driver` with environment variable to match?

mapkon:
> Just to provide a suggestion as a way to move forward:
> 
> Perhaps add a new configuration property to the config called `download_browser_driver` to choose to allow saml2aws to install the playwright browser for you with the new default going forward being not to?
> 
> Or maybe even a flag on `saml2aws login` called `--download-browser-driver` with environment variable to match?

I think this is a superb middle ground. Let's go with this!

pkg/provider/browser/browser_test.go
```diff
@@ -39,7 +39,8 @@ func TestValidate(t *testing.T) {
 	client, err := New(account)
 	assert.Nil(t, err)
 	loginDetails := &creds.LoginDetails{
-		URL: "https://google.com/",
+		URL:             "https://google.com/",
+		DownloadBrowser: true,
```
gliptak, line 43: is there a way to test fail without download?

pkg/provider/browser/browser.go
```diff
@@ -21,11 +21,21 @@ type Client struct {
...
-	return &Client{Headless: idpAccount.Headless}, nil
+	return &Client{
+		Headless: idpAccount.Headless,
+	}, nil
 }
 
 func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error) {
 
+	// Optionally download browser drivers if specified
+	if loginDetails.DownloadBrowser {
```
gliptak, line 32: will this (re)download/install at every run?

pkg/provider/browser/browser_test.go
xinkecf35: In a word, yes. It should be easy to copy/paste this test without the flag set and check if the error is not nil since I believe that's how the error is being propagated up to the user.

pkg/provider/browser/browser.go
xinkecf35: It shouldn't re-download since playwright-go does have that caching mechanism. Just by manually invoking the relevant `playwright install` command it doesn't re-download the binary, at least not without first nuking the relevant cache.

With this code, if you were to set it in the config file, it would functionally bring you back to the 2.36.4 behavior where it always checks if you have the drivers or not. If you do, it outputs the relevant message.

Alternatively, if you don't set it in the config, it will only download the driver if you set the relevant flag.

mapkon (approved):
@xinkecf35 / @gliptak 

I don't see anything holding this up. Are we good to go?

mapkon (commented):
Oh wait! I don't see the documentation updated. So that should be part of this

pkg/provider/browser/browser_test.go
xinkecf35: Okay, so I've added the test, though the change was largely than I had though it'll be. 

I had added a new config parameter to technically allow one to set the directory that Playwright installs into though I've purposely not documented.

I did this way because otherwise my test was going to be sensitive if the previous test ran and installed playwright before this test which introduces all sorts of headaches.

So instead, you technically can now choose where you download the drivers to, defaulting to the default location if you don't care.

xinkecf35: I left the existing test as it is to download the drivers to the default location  though an argument could be made it shouldn't.

mapkon:
@gliptak Thoughts?

mikex99:
Any way to pass a flag during 'saml2aws configure' that would set the 'download_browser_driver = true' to the config file?

gliptak:
@mikex99 I recommend targeting `saml2aws configure` in followup PR

gliptak (approved):

mikex99:
When will a new version with this merge be released?

mapkon:
> When will a new version with this merge be released?

We are organizing a test of the current codebase. Please see here: https://github.com/Versent/saml2aws/pull/1029#issuecomment-1533924283

If you can spend sometime to test out the current master and ascertain release readiness, I will appreciate

Issue #863, opened 2022-08-02
Make use of SHELL environment variable


everops-mat:
I typically use rc shell, but saml2aws exec uses /bin/sh -c '....' which causes some problems. I have a small PR that has exec pull the SHELL environment variable and us it if it exists, otherwise uses /bin/sh 

#862

gliptak:
an alternative approach might be to look up `/bin/bash` and fallback to `/bin/sh`

PR #1007, opened 2023-03-27
Implement /bin/sh option

gliptak:
#863

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ Commands:
     Emit a script that will export environment variables.
 
     -p, --profile=PROFILE      The AWS profile to save the temporary credentials. (env: SAML2AWS_PROFILE)
-        --shell=bash           Type of shell environment. Options include: bash, powershell, fish, env
+        --shell=bash           Type of shell environment. Options include: bash, /bin/sh, powershell, fish, env
         --credentials-file=CREDENTIALS-FILE
                                The file that will cache the credentials retrieved from AWS. When not specified, will use the default AWS credentials file location. (env: SAML2AWS_CREDENTIALS_FILE)
 
@@ -297,7 +297,7 @@ export AWS_CREDENTIAL_EXPIRATION="2016-09-04T38:27:00Z00:00"
 SAML2AWS_PROFILE=saml
 ```
 
-Powershell, and fish shells are supported as well.
+Powershell, sh and fish shells are supported as well.
 Env is useful for all AWS SDK compatible tools that can source an env file. It is a powerful combo with docker and the `--env-file` parameter.
 
 If you use `eval $(saml2aws script)` frequently, you may want to create a alias for it:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -277,7 +277,7 @@ Commands:
     Emit a script that will export environment variables.
 
     -p, --profile=PROFILE      The AWS profile to save the temporary credentials. (env: SAML2AWS_PROFILE)
-        --shell=bash           Type of shell environment. Options include: bash, powershell, fish, env
+        --shell=bash           Type of shell environment. Options include: bash, /bin/sh, powershell, fish, env
```
briantist, line 280: I don't use this part of `saml2aws` often, but just curious why the option would be named `/bin/sh` instead of `sh`. The other shell types don't take full paths.

gliptak: To ensure that system shell is invoked (regardless of PATH configuration)

I will move to `sh` if you think that is better

I didn't propose renaming /bin/bash and /bin/fish, is it might be breaking existing configs.

mapkon (commented):
I am wondering whether there is synergy with this PR: https://github.com/Versent/saml2aws/pull/862

mapkon:
> @mapkon I submitted this PR after viewing #862 I think the issue might be better addressed at the config level

OK! Should we close that PR then?

gliptak:
@everops-mat please comment

@mapkon this is an independent approach

mapkon (approved):

README.md
briantist: ok yeah I can see how that works from the invoking perspective; but the `script` subcommand does not actually invoke the shell it generates the variables for, no? I guess it's used for the shebang (though I can't see where it's used in this PR).

I might also argue that we _should_ respect the `PATH` variable, rather than deciding for the end user/sysadmin not to use what they configured.

Also looking at it from an end-user point of view, I think it's much nicer to refer to it as `sh`, both for consistency and just in general (I would not want to be expected to give the full path to a powershell executable for example, especially when it's not really needed to run this subcommand).

briantist: ah I see this has already been merged... if it has not been released, I would strongly urge reconsidering to use `sh` instead of the full path.

Perhaps, an additional option could be added to the `script` subcommand to provide a full path to the shell in question (the short name determines template, while the full path determines the binary).

This would also be nice for the `powershell` template since the binary can be named `powershell.exe`, `pwsh.exe`, or `pwsh`, for example.

briantist: @gliptak @mapkon pinging just because this conversation is marked "resolved", just didn't want the comments to get lost

gliptak:
@briantist thank you for your analysis. I didn't have a strong opinion on /bin/sh vs sh

Consider submitting a PR updating

mapkon:
> Sorry, this got buried. Loved the approach! Thanks for bring my idea to a reasonable implementation!

@everops-mat Can we close your PR the?

PR #1022, opened 2023-04-06
Add jumpcloud test

gliptak:


mapkon (approved):

mapkon (approved):

mcampbellsb:
Interestingly... related?  https://github.com/Versent/saml2aws/issues/1041

gliptak:
@mcampbellsb would you review https://github.com/Versent/saml2aws/pull/1022/files to check what I have had broke?

mapkon:
@mcampbellsb Could be. @gliptak is better placed to answer

mcampbellsb:
@gliptak @mapkon I'm not even close to qualified to debug, here.  I was hit by the XSRF issue yesterday and colleague noticed the issue reported on #1041 so we revered to an earlier version.  I was just going through PR's and noticed this one which looked to be at least in the same area.

If I were to _guess_, I don't think this PR broke anything, it might just be a missing test case, but again, that's just a guess.

PR #1047, opened 2023-05-12
Handling PingOne device names within button elements

rushndr:
In the environment I'm working in, the PingOne device names are within a 'button' element rather than an 'a' element. This change checks for device names within a 'button' element if there is no device name found within the 'a' element.


mapkon (commented):
@rushndr This looks fine. However, can you please add tests? This will reduce our test coverage

mapkon (approved):
LGTM - @gliptak?

wolfeidau:
@mapkon Looks good to me, great seeing more tests!

PR #1064, opened 2023-05-29
add another regex for finding state tokens

lizduty:
I've been having a weird intermittent issue, where trying to log in would sometimes fail with an opaque message: 

```
Error authenticating to IdP.: failed to getStateToken: error retrieving saml response: cannot find state token
```

After a bunch of digging, I discovered that sometimes Okta will request "extra verification" in the form of another 2fa request. Not a big deal, but for whatever reason that page doesn't store the state token as a single variable. Instead, it's burried deep in a blob of almost-JSON. 

This change lets us dig up state tokens from those pages as well, which mostly fixes the "cannot find state token" message


gempesaw:
woohoo thank goodness for this PR, i could only auth in once before getting this error and getting stuck. but now it's working like a charm, i can `login` as many times as i want :D

Issue #1078, opened 2023-06-14
Keycloak login loop Status="200 OK" on correct password


yeoldegrove:
This might be directly related to #948

We have a keycloak instance here with a LDAP user federation.
used: saml2aws v2.36.9 with keycloak 21.x.x

A few weeks ago we ran tests with keycloak 16.x.x and `saml2aws` worked fine there.
We see no login errors in our keycloak instance at all, so this is why I currently believe that this a bug in `saml2aws`.
We use an English locale (not a German one like in #948).

## Symptom

When running `saml2aws login` we land in an endless loop with an http status code of 200.

## Detailed logs

```
saml2aws login --verbose
DEBU[0000] Running                                       command=login
DEBU[0000] Check if creds exist.                         command=login
DEBU[0000] Expand                                        name=/home/user/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/home/user/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/home/user/.aws/credentials pkg=awsconfig
Using IdP Account default to access KeyCloak https://auth.acme.corp/auth/realms/acme/protocol/saml/clients/amazon-aws
To use saved password just hit enter.
? Username testuser3
? Password

DEBU[0000] building provider                             command=login idpAccount="account {\n  URL: https://auth.acme.corp/auth/realms/acme/protocol/saml/clients/amazon-aws\n  Username: testuser3\n  Provider: KeyCloak\n  MFA: Auto\n  SkipVerify: false\n  AmazonWebservicesURN: urn:amazon:webservices\n  SessionDuration: 3600\n  Profile: testuser3\n  RoleARN: \n  Region: \n}"
Authenticating as testuser3 ...

DEBU[0001] HTTP Req                                      URL="https://auth.acme.corp/auth/realms/acme/login-actions/authenticate?session_code=RG0GeIUru-_wo9QYqMUvsJBHO-M1HKwAxJJxPVu-WEI&execution=95ab09bf-f009-4d2a-a4bc-9c4c9162b0dd&client_id=amazon-aws&tab_id=LlxRTmyP2TE" http=client method=POST
DEBU[0001] HTTP Res                                      Status="200 OK" http=client
[...]

PR #1084, opened 2023-06-26
Ignore other login options

JanoschSchaefer:
Ignore other Keycloak login options. This will at least allow you to log in.

Relates to #1078

### DIFF
--- a/pkg/provider/keycloak/keycloak.go
+++ b/pkg/provider/keycloak/keycloak.go
@@ -14,12 +14,15 @@ import (
 	"github.com/PuerkitoBio/goquery"
 	"github.com/marshallbrekka/go-u2fhost"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/versent/saml2aws/v2/pkg/cfg"
 	"github.com/versent/saml2aws/v2/pkg/creds"
 	"github.com/versent/saml2aws/v2/pkg/prompter"
 	"github.com/versent/saml2aws/v2/pkg/provider"
 	"github.com/versent/saml2aws/v2/pkg/provider/okta"
 )
+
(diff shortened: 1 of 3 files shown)


pkg/provider/keycloak/keycloak.go
```diff
@@ -393,6 +393,8 @@ func updateKeyCloakFormData(authForm url.Values, s *goquery.Selection, user *cre
 		authForm.Add(name, user.Username)
 	} else if strings.Contains(lname, "password") {
 		authForm.Add(name, user.Password)
+	} else if strings.Contains(lname, "tryanotherway") {
```
mapkon, line 396: If the intention is to ignore the case, maybe we should log that as the intention?

mapkon (commented):
See my comment

pkg/provider/keycloak/keycloak.go
JanoschSchaefer: I think logging the information will produce more noice than provide any relevant information to the user. Therefore i haven't added any handling besides ignoring this case.

yeoldegrove: If the "Try another Way" button is present, the keycloak admin has configured other methods to login.
These are effectively ignored when adding this code, which is totally fine for now (in my opinion).
Nevertheless, the ultimate goal should be adding the possibility to use the other login methods.
Therefore, a small message like: "Ignoring other ways to log in (not implemented)" would be great... Also for the next person who has to debug this ;)

JanoschSchaefer: Added debug log that other options are currently not implemented

mapkon (commented):
Looks good! @JanoschSchaefer / @yeoldegrove - I am happy to merge this if you are?

PR #1119, opened 2023-08-31
Use logger instead of fmt.Println in brwoser.go

alFReD-NSH:
`fmt.Println` still logs when `--quiet` is specified and that's problematic when saml2login is being used as credential_process in aws profiles.

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -2,7 +2,6 @@ package browser
 
 import (
 	"errors"
-	"fmt"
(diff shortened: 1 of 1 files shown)


mapkon (approved):

mapkon (changes requested):
Please look into the lint errors

alFReD-NSH:
@mapkon Thanks for the quick review, fixed the lint error.

mapkon (approved):

Issue #1079, opened 2023-06-19
panic: runtime error: index out of range [1] with length 1


Devansh2290:
I have configured `saml2aws` on my new macbook pro M1 chip, and when trying to login into `aws` using following command getting this error:

command:
```
saml2aws login --session-duration=43200 --username=<email id> --mfa='DUO' --duo-mfa-option='Duo Push' --skip-prompt  --force --role='arn:aws:iam::<account>:<role>'
```

Error:
```
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/versent/saml2aws/v2/pkg/provider/okta.verifyMfa(0x140003f4cc0, {0x140003de2d9, 0xf}, 0x140001584d0, {0x1400042c380, 0x336})
        github.com/versent/saml2aws/v2/pkg/provider/okta/okta.go:975 +0x6dfc
github.com/versent/saml2aws/v2/pkg/provider/okta.(*Client).Authenticate(0x140003f4cc0, 0x140001584d0)
        github.com/versent/saml2aws/v2/pkg/provider/okta/okta.go:481 +0x184
github.com/versent/saml2aws/v2/pkg/provider/okta.(*Client).authWithSession(0x140003f4cc0, 0x140001584d0)
        github.com/versent/saml2aws/v2/pkg/provider/okta/okta.go:300 +0xbc
github.com/versent/saml2aws/v2/pkg/provider/okta.(*Client).Authenticate(0x140003f4cc0, 0x140001584d0)
        github.com/versent/saml2aws/v2/pkg/provider/okta/okta.go:463 +0x54
github.com/versent/saml2aws/v2/cmd/saml2aws/commands.Login(0x14000154cc0)
        github.com/versent/saml2aws/v2/cmd/saml2aws/commands/login.go:105 +0x37c
main.main()
        github.com/versent/saml2aws/v2/cmd/saml2aws/main.go:191 +0x7240
```

`$ saml2aws --version`
2.36.8

hmoghani:
Seeing the same issue on Mac Intel. I have the latest version: 2.36.10

`panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/versent/saml2aws/v2/pkg/provider/okta.verifyMfa(0xc00047aed0, {0xc0004663c9, 0x13}, 0xc0001ce4d0, {0xc0000f0380, 0x36b})
[...]

Issue #1061, opened 2023-05-25
Array index out of bounds error with Okta and MFA


simontrasler:
When running saml2aws 2.36.8, I get the output included below. I am able to authenticate with no issues using version 2.36.5. Let me know if you need more information.

I suspect this is an issue with the new block of code https://github.com/Versent/saml2aws/blob/v2.36.8/pkg/provider/okta/okta.go#L947-L973. This builds an array from one data source, and an index into that array from an another source, and expects them to match. In my case, the array length is 1, and the wanted index into that array is 1. The result is a crash.

```
strasler@strasler-mbp work % saml2aws --verbose login --duo-mfa-option 'Duo Push'
DEBU[0000] Running                                       command=login
DEBU[0000] Check if creds exist.                         command=login
DEBU[0000] Expand                                        name=/Users/strasler/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/Users/strasler/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/Users/strasler/.aws/credentials pkg=awsconfig
Using IdP Account default to access Okta https://springserve.okta.com/home/amazon_aws/XXX/272
DEBU[0000] Get credentials                               helper=osxkeychain serverURL="https://springserve.okta.com/home/amazon_aws/XXX/272"
To use saved password just hit enter.
? Username strasler
? Password ***********************

DEBU[0032] building provider                             command=login idpAccount="account {\n  DisableSessions: false\n  DisableRememberDevice: false\n  URL: https://springserve.okta.com/home/amazon_aws/XXX/272\n  Username: strasler\n  Provider: Okta\n  MFA: DUO\n  SkipVerify: false\n  AmazonWebservicesURN: urn:amazon:webservices\n  SessionDuration: 43200\n  Profile: default\n  RoleARN: arn:aws:iam::XXX:role/SSO-Developer\n  Region: us-east-1\n}"
DEBU[0032] okta | disableSessions: false                 provider=okta
DEBU[0032] okta | rememberDevice: true                   provider=okta
Authenticating as strasler ...
DEBU[0032] HTTP Req                                      URL="https://springserve.okta.com/api/v1/authn" http=client method=POST
DEBU[0033] HTTP Res                                      Status="200 OK" http=client
[...]

PR #1133, opened 2023-09-21
fix: find the chosen okta mfa option in the mfaOptions

russell:
find the index of the a duoMfaOption by looking in the duoMfaOptions.

the duoMfaOptions is dynamically created, so it could have any number of items, the reason that using the interactively  selected mechanism works is that it's choosing based on the list of duoMfaOptions options rather than hardcoded ids.  This change does something similar, but searches through duoMfaOptions to find the option passed at the CLI, if the item is missing, then an error will be emitted.

for example if I don't have passcode enabled and if i try to force that option
```
$ ./saml2aws login --session-duration=43200 --mfa='DUO' --duo-mfa-option='Passcode' --skip-prompt --force --role='arn:aws:iam::183823948234:role/okta/team'
Using IdP Account default to access Okta https://foobar.okta.com/app/amazon_aws/aoeusnthoaenuth/sso/saml
Authenticating as rsim@foobar.com ...
Error authenticating to IdP.: error verifying MFA: error unsupported MFA option 'Passcode'
```

fixes https://github.com/Versent/saml2aws/issues/1079 https://github.com/Versent/saml2aws/issues/1061


mapkon (changes requested):
Hey - do you mind adding tests to this? This is going to push the coverage down and make the build to fail

russell:
Sure! here is a test, can you give me some feedback about the fixtures, i haven't trimmed them down to only the fields that are referenced in the okta.go calls, if you like i can do that.  I left them with more information rather than less in the hope that future readers would be able to improve the logic without needing access to fresh raw responses from duo.  But each to their own, would you like them trimmed?

mapkon (approved):

mapkon:
> Sure! here is a test. Can you give me some feedback about the fixtures, i haven't trimmed them down to only the fields that are referenced in the okta.go calls, if you like i can do that. I left them with more information rather than less in the hope that future readers would be able to improve the logic without needing access to fresh raw responses from duo. But each to their own, would you like them trimmed?

I personally think we should trim them down so that the test is focused on what it is testing. The extra context can be confusing.

@gliptak thoughts?

russell:
> I personally think we should trim them down so that the test is focused on what it is testing. The extra context can be confusing.

i'm cool with doing that, i'll update the PR later today.  I'm a bit on the fence personally so it's not a big deal

russell:
ok, this is ready for review again, I have extended the test suite to assert all the server request's and trimmed down the fixtures

mapkon (approved):

mapkon:
@russell  This has been released

Issue #1137, opened 2023-10-04
SMS never sent when login with GoogleApp provider


ksilverwall:
## Problem

SMS never sent when login with GoogleApp provider.
(I am not hitting the SMS sending limit because other tool success to send SMS)

```
$ saml2aws login
Using IdP Account default to access GoogleApps https://accounts.google.com/o/saml2/initsso****
To use saved password just hit enter.
? Username kazuki.tsunemi@timee.co.jp
? Password ***********

Authenticating as kazuki.tsunemi@timee.co.jp ...
? Enter SMS token: G-
```

## Settings

`.saml2aws`

```
[default]
...
provider                = GoogleApps
mfa                     = Auto
...
```

ksilverwall:
is this related issue?
https://github.com/cevoaustralia/aws-google-auth/issues/195

edwardrf:
I am experiencing the same issue, using `DUMP_CONTENT=true` shows google is showing a page with

```
This extra step shows it’s really you trying to sign in
Get a verification code by text message at: (•••) •••-••01
Send text message
Don’t ask again on this device
For your convenience, keep this checked. On shared devices, additional precautions are recommended.
Learn more
```

Which needs click on Send text message.

PR #1150, opened 2023-10-19
Handle extra "Send text message" challenge page

edwardrf:
It seems some times google will present an extra "Send text message" challenge page that requires the user to click on the button to actually get text message sent and present the code input in the next page.

#1137

### DIFF
--- a/pkg/provider/googleapps/example/challenge-sms-send.html
+++ b/pkg/provider/googleapps/example/challenge-sms-send.html
@@ -1,0 +1,350 @@
+<!doctype html>
+<html lang="en" dir="ltr">
+
+<body id="yDmH0d">
+  <div class="s2h6df">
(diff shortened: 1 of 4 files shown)


lionello:
@mapkon  can you check?

mapkon:
> @mapkon can you check?

@edwardrf Thank you for these changes. Do you mind adding some tests to this?

edwardrf:
@mapkon sorry for the delay, just added the test case, plz take a look.

edwardrf:
@mapkon i've updated the test case that wasn't functioning correctly due to automatic html formatting, plz take a look again.

mapkon:
> @mapkon i've updated the test case that wasn't functioning correctly due to automatic html formatting, plz take a look again.

Thanks buddy! I will take a look

PR #1153, opened 2023-10-26
Add 5m timeout to page.WaitForRequest

kbarlowgw:
I didn't see an appropriate place to add/update a unit test for this but I have been using it locally and it fixes the issue.  Doesn't seem like an external configuration flag is required but may be if users continue to have issues.

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -13,6 +13,8 @@ import (
 
 var logger = logrus.WithField("provider", "browser")
 
+var TimeoutPageWaitForRequest = newPageWaitForRequestOptions(300000)
+
 // Client client for browser based Identity Provider
 type Client struct {
 	Headless bool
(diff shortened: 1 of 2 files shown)


mapkon:
@kbarlowgw How about `browser_test.go`?

I think you need find a way for this path `playwright.PageWaitForRequestOptions{Timeout: &timeout}` to be exercised so that test coverage is not affected.

kbarlowgw:
Thanks... I pushed an update that will hopefully fix it.  It went from
46.7% to 48.9% locally:

ok   github.com/versent/saml2aws/v2/pkg/provider/browser 1.935s coverage:
48.9% of statements

Please let me know if there's anything else required.

Keith

mapkon (approved):

alediaferia:
Any chance this could be made configurable via e.g. command line switch?

kbarlowgw:
I can look into that.

Keith

kbarlowgw:
It looks to me that the existing timeout field in the account settings was
unused in browser configurations so I mapped that to the timeout for
Page.WaitForRequest.  If the timeout is set less than 30s, the default of 5
mins will be used.  It seems pointless to allow timeouts less than the
default provided by playwright.  Unit tests have been added & updated.

Keith

kbarlowgw:
Thanks, all!

Keith

dkimdon:
This change works great for me.  What needs to happen to get this in the next release?

mapkon:
I just need to cut a release. I am looking to a window to do that

Issue #1161, opened 2023-11-09
Unable to login using SAML 2.0 federated users to access the AWS Management Console due to recent change.


eliat123:
A new code change that was introduced 5 days ago in https://github.com/Versent/saml2aws/pull/1157 is causing the following bug to occur:

**Background**
We are using "SAML 2.0 federated users to access the AWS Management Console" as described in: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_enable-console-saml.html

**Symptom**
After the identity verification, the browser gets redirected to "https://signin.aws.amazon.com/saml" which immediately gets redirected to "about:blank", the user should have been prompted for role selection. However, the login process does not complete because the user is not able to select a role in the UI.

For your reverence, here is step 7 from AWS docs:
```
7. The client browser is redirected to the AWS Management Console. 
If the SAML authentication response includes attributes that map to multiple IAM roles, 
the user is first prompted to select the role for accessing the console.
```
The expected page is described at https://aws.amazon.com/blogs/security/easier-role-selection-for-saml-based-single-sign-on/

The interaction with the browser provider used to work until this commit was merged.

**Proposed solution**
It can be fixed if you change the following lines in pkg/provider/browser/browser.go change:
```
r, _ := page.ExpectRequest(signin_re, func() error {
		_, err := page.Goto("about:blank")
		return err
	}, client.expectRequestTimeout())
```
To: 
```
r, _ := page.ExpectRequest(signin_re, func() error {
		return nil
	}, client.expectRequestTimeout())
```

eliat123:
Hi @gliptak, 
Can you please share what was the reason for the redirect to "about:blank" in https://github.com/Versent/saml2aws/pull/1157/files#diff-5cd59a5a8d31c61bb9c035ee3524699cd8e97b722683501d805074f7d11b7ee9R102-R103
This is causing a bug where `saml2aws` does not prompt for MFA code but being redirected immediatly.

Is this mandatory? 
Can we not redirect in some cases?

gliptak:
there was a signature change with new release and this method has an `error` callback:

https://github.com/playwright-community/playwright-go/blob/main/page.go#L567

the blank page likely indicates that there was a failure

feel free to submit PR improving

PR #1157, opened 2023-11-04
Bump github.com/playwright-community/playwright-go from 0.3500.0 to 0.3900.0

gliptak:
Different operation used, resulting executable needs to be tested

Mock generation example running within `make docker-build-environment` 

````
go get github.com/vektra/mockery/v2@v2.36.0
.buildtemp/bin/mockery --output mocks/ --dir .buildtemp/pkg/mod/github.com/playwright-community/playwright-go\@v0.3900.0/ --name Response
````

#1122

PR #1164, opened 2023-11-14
[Bug fix] remove redirect to about:blank page

eliat123:
Solves #1161

### Background:
The code change that was introduced in https://github.com/Versent/saml2aws/pull/1157/files#diff-5cd59a5a8d31c61bb9c035ee3524699cd8e97b722683501d805074f7d11b7ee9R102-R103

was done because of a signature change with the new release:
https://github.com/playwright-community/playwright-go/blob/main/page.go#L567

The `cb` method is not an error callback, but a general callback which is not mandatory:

https://github.com/playwright-community/playwright-go/blob/ddc7abdd06db016d63664de1ab08870195085ddb/waiter.go#L144

### The issue
The redirect to blank page created an unwanted side-effect that did not allow saml2aws to continue with AWS role selection as described in #1161.

This is the fix.

FYI @gliptak

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -141,10 +141,7 @@ var getSAMLResponse = func(page playwright.Page, loginDetails *creds.LoginDetail
 	}
 
 	logger.Info("waiting ...")
-	r, _ := page.ExpectRequest(signin_re, func() error {
-		_, err := page.Goto("about:blank")
-		return err
-	}, client.expectRequestTimeout())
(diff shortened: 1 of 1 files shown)


gliptak (approved):
Thank you for correcting my understanding of updated behaviour and working through this @eliat123

mapkon (approved):

mapkon:
CC: @gliptak

gliptak:
:+1:

PR #1183, opened 2023-12-14
Add browser autofill

### LINKED ISSUE #1182: Browser autofill
Would be nice if the browser provider could autofill the username and password that we provide earlier. Password field is easy to find and username would the other input field around it, something simple behind a flag would save so much time of people typing in passwords!

I'm working on this to make a PR.

faridnsh:
Fixes #1182

We could make the code a little bit smarter, or even have the field css selectors(username, password, submit button) configured from config file instead.


tinaboyce:
Hi @alFReD-NSH ,

Thanks for doing a PR. Are you able to provide tests for this?

alFReD-NSH:
There's no tests currently for the browser provider, I'm not sure what's the reason for it, but I'm guessing it's because it's not as easy as other providers. I would appreciate if some directions or tips could be provided. I'll give it a try in couple of weeks.

mapkon:
> There's no tests currently for the browser provider, I'm not sure what's the reason for it, but I'm guessing it's because it's not as easy as other providers. I would appreciate if some directions or tips could be provided. I'll give it a try in couple of weeks.

Can you reference `browser_test.go`?

alFReD-NSH:
@mapkon

I did indeed have a look at it and the actual part that runs the browser provider for a successful outcome is commented out(line 159). I tried to write a similar test with the mocked page type that already exists and I get the following error:

```
pkg/provider/browser/browser_test.go:249:18: cannot use page (variable of type *mocks.Page) as playwright.Page value in argument to autoFill: *mocks.Page does not implement playwright.Page (wrong type for method ExposeFunction)
                have ExposeFunction(string, func([]interface{}) interface{}) error
                want ExposeFunction(string, func(...interface{}) interface{}) error
```

I tried regenerating the mocks, but didn't work, Mockery doesn't detect the difference. This could be fixable by Mockery, so I'll report an issue there.

alFReD-NSH:
Apparently it's a known issue and they have a PR for it: vektra/mockery#719

mapkon:
OK - I will review this and merge it

tinaboyce:
Hi all (paging @mapkon), I have updated the PR to include a test. I have managed to write test utilising a server and running the autoFill function against a hosted example form page.

mapkon (approved):
This is brilliant @tinaboyce - I will take a look later

Issue #912, opened 2022-11-15
Add support for Okta FIDO2 biometric authentication


seppestas:
Okta supports [FIDO2 platform authentication that's integrated into a device and uses biometric data](https://help.okta.com/en-us/Content/Topics/Security/mfa-webauthn.htm), such as Windows Hello or Apple Touch ID.

I'm able to use a Windows Hello fingerprint scanner on the Okta web app, which is awesome. It would be great if saml2aws also support it so I can stop having to use my phone for authentication. I can imagine supporting Apple Touch ID would also help many people.

Related to #419, though that issue talks about YubiKeys, which seem to be supported already.

seppestas:
Note: it seems like the WebAuthn method is found by saml2aws, however it does not work. When running `saml2aws login` with MFA method set to Auto and the Windows biometric fingerprint reader registered to my Okta account I get presented the following MFA options:

? Select which MFA option to use FIDO:
WebAuthn MFA authentication
Okta MFA authentication
PUSH MFA authentication

If the biometric MFA factor is not registered, I do not get the option.

However, when I try to log in, I get the following error:
error authenticating to IdP: error verifying MFA: error retrieving verify response: request for url:
```
https://<redacted>.okta.com/api/v1/authn/factors/<redacted>/verify failed status: 401 Unauthorized
```

Running with the `--verbose` flag also shows following error:
```
no U2F devices found. device might not be plugged in
error verifying MFA
```

So it looks like the underlying go-u2fhost does not support Windows Hello U2F devices.

jwshieldsGetty:
I believe I am running into a similar issue, receiving the same error message regarding "no U2F devices found" -  
but I **do not** use Windows Hello.  

I would say the only thing is that my Yubikey (5C, not a Bio) has a PIN set for U2F/FIDO auth. Which makes me wonder if the PIN is the issue here, as it was pointed out in this comment: https://github.com/Versent/saml2aws/issues/419#issuecomment-741916350  

@seppestas do you have a PIN enabled on your Yubikey device?

seppestas:
I am not using a Yubikey, but a fingerprint reader built into my laptop. This provides 2FA through the Windows hello framework, which supports WebAuthn (FIDO2), not FIDO U2F.

Your Yubikey 5C should work, since it supports FIDO U2F. Maybe create a ticket in https://github.com/marshallbrekka/go-u2fhost if it doesn't or ensure FIDO U2F is enabled.

[...]

PR #1221, opened 2024-02-23
feat: Add support for Windows WebAuthn 2FA

seppestas:
See #912. This uses a new library I created to use system level WebAuthn client libraries. Currently this only supports Windows Hello, I hope to one day also support MacOS Touch ID.

Note: I made some last minute changes before creating this PR that I was not able to fully verify due to me not being able to remove sessions / force 2FA.


mapkon (changes requested):
The build is failing, can you take a look?

seppestas:
@mapkon I had a look, but I think this might just be a regression in the test itself? Or is it just failing to build for ARM windows?

The error is ["failed to build for windows_arm_6: exit status 1: go: downloading github.com/sirupsen/logrus v1.9.3"](https://github.com/Versent/saml2aws/actions/runs/8023005142/job/21931587293?pr=1221#step:5:41).

This sounds to me like it might be failing to download a different package (logrus), or does goreleaser just print the last thing it did for some reason?

If there is a problem, I would expect it to be with building for ARM Windows, which is something I did not test TBH. Can I somehow get the logs of the go build?

seppestas:
Turns out the problem is that CGO is not supported on Windows ARM, and somehow this just causes the whole webauthn_windows.go to be ignored iof. showing a useful error message. Even worse is that I can compile the package just fine, only when compiling the test / saml2aws it fails, because there is no explicit API :/.

Go-u2fhost, i.e [bearsh/hid](https://github.com/bearsh/hid) solves this by adding a !cgo build constraint, I'll do the same for now. This means using Windows Hello for 2FA will not be supported (nor is it when using raw USB authenticators).

In theory I can remove the dependency on cgo, I only use it as an easy way to create the structs I need from the C header files. In the long run, I would like to use something like [winrt](https://github.com/saltosystems/winrt-go), but I am not sure if this is possible.

seppestas:
@mapkon it should work with the latest changes.

mapkon:
> @mapkon it should work with the latest changes.

I will take a look

mapkon (approved):

mapkon:
@seppestas This looks fine. Any reason why you omitted tests?

seppestas:
Eh, there's not all that much to test. I suppose error handling could be tested?

mapkon:
> Eh, there's not all that much to test. I suppose error handling could be tested?

Yeah - the codecov report above could help. Its not really mandatory, but I like to cover these bases since it makes future changes easier to merge, and regressions easier to spot.

seppestas:
I had a look at adding tests. The most useful test would be to test `func fidoWebAuthn`, i.e. verify that if no `NewFidoClient` is found (or any other error occurs), it will try to use the system Webauthn library instead.

Please also help in thinking critically about this. This is a scenario that works for me, but it might not work for others. E.g people with a Yubikey, that want to use their Yubikey through Windows Hello, might not be happy with this solution. See [my comment in #912](https://github.com/Versent/saml2aws/issues/912#issuecomment-1874089606). I think the "fall back to system Webauthn if no U2F USB device found" approach is good enough for most cases.

Sadly, no tests exists currently for the `fidoWebAuthn` function. The only tests that are there are for `ChallengeU2F`, and it only really tests the successful condition, arguably also quite useless, while there is plenty more interesting stuff like timeouts and error conditions.

I'm happy to write tests for the goal of improving quality, but not just to get a coverage number go up. To make the coverage number go up (or at least, to cover most of the lines I added), I could write a (arguably useless) test for `ChallengeSystemWebAuthn` that would do little more than verifying the contents of the `SignedAssertion` and checks if the error is propagated. This test would also be non-trivial, because I do not use any interfaces / objects, I just use a single function.

mapkon:
> I had a look at adding tests. The most useful test would be to test `func fidoWebAuthn`, i.e. verify that if no `NewFidoClient` is found (or any other error occurs), it will try to use the system Webauthn library instead.
> 
> Please also help in thinking critically about this. This is a scenario that works for me, but it might not work for others. E.g people with a Yubikey, that want to use their Yubikey through Windows Hello, might not be happy with this solution. See [my comment in #912](https://github.com/Versent/saml2aws/issues/912#issuecomment-1874089606). I think the "fall back to system Webauthn if no U2F USB device found" approach is good enough for most cases.
> 
> Sadly, no tests exists currently for the `fidoWebAuthn` function. The only tests that are there are for `ChallengeU2F`, and it only really tests the successful condition, arguably also quite useless, while there is plenty more interesting stuff like timeouts and error conditions.
> 
> I'm happy to write tests for the goal of improving quality, but not just to get a coverage number go up. To make the coverage number go up (or at least, to cover most of the lines I added), I could write a (arguably useless) test for `ChallengeSystemWebAuthn` that would do little more than verifying the contents of the `SignedAssertion` and checks if the error is propagated. This test would also be non-trivial, because I do not use any interfaces / objects, I just use a single function.

Makes sense! You make a sound point so I am happy to merge either way. Do you want to write the test or should I just go ahead and merge?

seppestas:
@mapkon thanks for merging this. If I find some spare time I might add some tests. I am hoping to add support for Linux and MacOS sometime in the future.

There's something to be said in at least making the Authn library more test friendly. I'll probably change the API in something that provides a list of authenticators, one of which could be e.g. Windows Hello. This is probably also a bit more elegant than returning an error on unsupported platforms, and it would allow platforms like Linux to have multiple options.

Any idea about how many users would use Windows on ARM? Would this be significant enough to try and get rid of CGO?

Either way, I have been using this feature daily since January, I hope other people find it useful as well.

mapkon:
That is an excellent question (re: windows user base). However, I have no idea how many users consume the app on windows ARM. I know for sure that. @briantist is a windows advocate around these circles.

*Re: CGO*

From  I think @wolfeidau has always wanted to get rid of CGO, but we never got around to it. There is a case to be made for simplifying the project by removing the CGO dependencies.

CC: @gliptak

seppestas:
Re: CGO

I'll have a look. I did not really consider CGO would be this much of a pain. The worst part is that I started out creating my library without CGO, but only later added it to easily re-use the C structures in the official Windows WebAuthn API headers.

In an even more perfect world, I would be able to get it working Windows COM / WinRT, making it properly async. [I looked into this](https://github.com/saltosystems/winrt-go/issues/78), but I don't think it is supported, though in theory possible using lower level  or maybe the Smartcard WebRT libraries.

mapkon:
Most definitely (and thanks for looking into this). @wolfeidau do you have any input/preference on this?

PR #1222, opened 2024-02-24
Okta + Duo: Allow passcodes from mobile app

wlonkly:
Any Duo mobile app registered with Duo, usually for Duo Push, also has the capability to produce HOTP passcodes:

Duo Mobile app on iOS, showing HOTP code. Yes, I burned that code. 

In the case where Duo Push is administratively disabled and Yubikeys (etc) are not provided, HOTP might be the only way for a user to perform Duo MFA. This is the case for us, which is preventing our non-Yubikey users from using saml2aws at all. 

This change adds Passcode to the list of MFA possibilities whenever a `phone1` is registered with Duo, to support using HOTP. 

--- 

I debated between adding it like this, or adding another conditional to the `option[value="token"]` case -- let me know if that would be preferable. I also checked the other Duo-enabled providers, but none of the others use this kind of logic to limit the MFA options presented to the user.


mapkon (approved):
Looks fine to me. @gliptak / @wolfeidau - any thoughts?

gliptak:
any related unit test updates?

gliptak:
@wlonky yes, there seems to be coverage the modified code https://app.codecov.io/gh/Versent/saml2aws/blob/master/pkg%2Fprovider%2Fokta%2Fokta.go#L974

consider creating a copy of https://github.com/Versent/saml2aws/blob/14c45abf5f8dcee4d39a269cf1a55dcf28006643/pkg/provider/okta/okta_test.go#L268 and updating for 1Passcode` path (some processing could be shared between these two flows)

wlonkly:
Got it, I'll see what I can do! Probably won't get to this until the weekend.

wlonkly:
Sorted and ready to go!

pkg/provider/okta/okta_test.go
```diff
@@ -266,9 +268,108 @@ func TestVerifyMfa(t *testing.T) {
...
+	loginDetails := &creds.LoginDetails{URL: ts.URL, Username: "user@example.com", Password: "test123"}
+	return ac, loginDetails
+}
+
+func setupTestDuoHttpServer(t *testing.T, duoFactor string) (*httptest.Server) {
 	verifyCounter := 0
 	statusCounter := 0
-	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+  ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
gliptak, line 372: this seems to have wrong indent

wlonkly: ty, fixed

mapkon:
@gliptak Is this ready to merge?

mapkon (changes requested):
Moving back to needs work since there is an issue with Okta

mapkon:
> Rats, I need to make one more change -- I just used my local build on my actual work Okta account, and I see duplicate "Passcode" options (because I have both `token` and `phone1` authenticators; my test account only had `phone1`).

OK - just let me know when its ready

wlonkly:
@mapkon Should be all set now. I switched around the logic to only add `Passcode` once, rather than uniq-ing it at the end.

mapkon (approved):

PR #1242, opened 2024-03-21
Browser provider - Autofill works with more authentication forms

vjeantet:
Hello, 

This PR proposition updates the way the autofill func submits the form:
Instead of locating the submit button and clicking on it, it send an Enter keypress.

I had to implement this to make it works with my company authentification web form, which has no html submit button, but an image with javascript events linked to it.

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -192,9 +192,7 @@ var autoFill = func(page playwright.Page, loginDetails *creds.LoginDetails) erro
 		return err
 	}
 
-	//If the submit button exists in the page, click it
-	// Press Enter to send the current form
-	// Find the submit button of the form that the password field is in
+	// Find the submit button or input of the form that the password field is in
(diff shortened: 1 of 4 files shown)


mapkon (commented):
Thank you for this PR.

However, aren't we making the implicit assumption that the page will have a submit button as a default or primary that responds to the enter click event? What happens if it is not?

vjeantet:
Hello, That won't work.

I can suggest something that covers all cases:

WHEN the input[type="submit"] exists, THEN click on it.
WHEN the input[type="submit"] does not exist THEN press ENTER.

mapkon:
> Hello, That won't work.
> 
> I can suggest something that covers all cases:
> 
> WHEN the input[type="submit"] exists, THEN click on it. WHEN the input[type="submit"] does not exist THEN press ENTER.

Sounds much more resilient. If you don't mind, please add some tests to hasten the reviews.

vjeantet:
updated to handle the two cases, with tests.

mapkon (approved):

Issue #1259, opened 2024-04-30
Errors using GoogleApps - error loading first page


aturbenson:
When using the GoogleApps provider with saml2aws, I'm getting a error: 

 ```
Error authenticating to IdP.: error loading first page: failed to build login form data: could not find any forms matching the provided IDs An error occurred
```

[Looking at the code](https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L269) I see that saml2aws is looking for a form with `id=gaia_loginform` , but the html I get when running saml2aws login doesn't have that element. It looks like google might have made an update to their login form so the form can no longer be accessed with that id. 

Is there a workaround so saml2aws login can work with the updated Google form?

**Comment by mjhoffman65** [comment]
Looks like aws-google-auth is having issues as well due to Google's redesigned login UI. There's a [proposed fix](https://github.com/cevoaustralia/aws-google-auth/pull/281) that we can use as a reference.

**Comment by jmctune** [comment]
The workaround right now is to log in via the browser method until someone patches it:

```
SAML2AWS_IDP_PROVIDER=Browser saml2aws login --download-browser-driver
```

austinfinnql:
> The workaround right now is to log in via the browser method until someone patches it:
> 
> ```
> SAML2AWS_IDP_PROVIDER=Browser saml2aws login --download-browser-driver
> ```

The ^ command didn't work for me. In case anyone else has a similar issue, you can try appending the account name to command . The full command that worked for me ended up being `SAML2AWS_IDP_PROVIDER=Browser saml2aws login --download-browser-driver -a uat`

aaronthebaron:
I have a pull request open to fix this issue as it's impacting most of the users in my org. I have been able to test for myself the new login flow with a 2FA phone app.

If any of you have a different 2FA you want to test this code against, it would be appreciated.

You can build my branch and test with the local copy:
https://github.com/aaronthebaron/saml2aws/tree/master?tab=readme-ov-file#building

edan-shalom:
Same here.. Thanks for sharing! Will give it a try as well.

aaronthebaron:
If you find it's not working, run the command with `DUMP_CONTENT=true` and `--verbose` flag. You can give me that output in the PR and I'll see if I can fix. Thanks!

vtstanescu:
I suppose this can be closed with the merging of https://github.com/Versent/saml2aws/pull/1285

mjhoffman65:
[...]

PR #1285, opened 2024-06-05
Update the googleapps IDP provider to work with changes to Google log…

aaronthebaron:
…in page while maintaining gaialogin compatibility

Attempting to fix the issues with the GoogleApps IdP provider noted in the issue #1259 

The google login system has changed for most but not all users causing this provider to no longer work for a majority of our in house users.

I have tested this with the new pages, in a configuration requiring 2-factor authentication with a phone app. I have not been able to regression test this with the current login, but I have tried to keep that legacy flow intact with my changes. I have also not been able to test each challenge but I have modified the challenges in a way that I hope to be backwards compatible.

### DIFF
--- a/pkg/provider/googleapps/example/captcha.html
+++ b/pkg/provider/googleapps/example/captcha.html
@@ -1,0 +1,102 @@
+<!doctype html>
+<html lang="en" dir="ltr">
+  <head>
+    <title>Sign in - Google Accounts</title>
+  </head>
+  <body>
+    <div class="BDEI9 LZgQXe">
+      <div class="Ha17qf" data-auto-init="Card">
+        <div class="Or16q">
+          <div data-view-id="hm18Ec" data-locale="en" data-allow-sign-up-types="true"><c-wiz jsrenderer="OTcFib" jsshadow jsdata="deferred-i2" data-p="%.@.]" data-node-index="2;0" jsmodel="hc6Ubd" c-wiz>
+              <div class="gEc4r"><img src="//ssl.gstatic.com/images/branding/googlelogo/2x/googlelogo_color_74x24dp.png" class="TrZEUc" alt="Google" width="74" height="24"></div>
+              <div class="EQIoSc" jsname="bN97Pc">
+                <div jsname="paFcre">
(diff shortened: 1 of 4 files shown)


edwardrf:
Awesome PR, but it seems I am still unable to login with the following error:
```
Using IdP Account default to access GoogleApps https://accounts.google.com/o/saml2/initsso?idpid=C00mmafdv&spid=XXXXXXXXXXXX&forceauthn=false
Authenticating as XXX@XXXXXX.XXX ...
Error authenticating to IdP.: error loading challenge page: unable to extract skip form: could not find form with query "form[action$=\"skip\"]"
```

mapkon (commented):
@aaronthebaron is this ready for review?

Also, lots of changes here - do you reckon we should cover some of the missing lines with tests?

edwardrf:
@aaronthebaron 
Please see the log in the attachment
[test.log](https://github.com/user-attachments/files/15603282/test.log)

aaronthebaron:
@mapkon 

> @aaronthebaron is this ready for review?
> 
> Also, lots of changes here - do you reckon we should cover some of the missing lines with tests?

Yes, this is ready for review.

Of all the lines I've added, most of them are under existing testing. I've added a test and modified another test to cover the two main changes.

There are so many paths possible in this provider that I don't have access to, but current this is passing regression.

More paths will be breaking that I won't be able to fix and should probably be handed in another pull reqest.

aaronthebaron:
@edwardrf Thank you for providing the output. Unfortunately you are entering a part of the flow that handles multiple challenge possibilities (TOTP, U2F, etc) and that page has completely changed.

Originally it must've been a form with multiple actions but now looks to be divs with different annotations. I think a lot more work will need to happen to get this page working again and I'm not sure I can spend more time on it.

The code for that page can be found [here](https://github.com/Versent/saml2aws/blob/master/pkg/provider/googleapps/googleapps.go#L497).

A quick fix for you personally might be to remove some possible authentication methods. Sorry I can't do more at this time.

mapkon (approved):

PR #1286, opened 2024-06-06
Challenge selection fix

aaronthebaron:


aaronthebaron:
@edwardrf I don't know if this helps you at all, but in this branch I put some basic work into fixing the issue you're running into with change of selection.

When you enter googleapps.go:83 you're failing out to the bottom at googleapps.go:449

Instead, I think we need to catch the error and try to look at this selection page, so in this branch I added a new function and an if block to try and catch your new selection page.

The new selection page is the html that is added to tests in this PR. I think the key is figuring out which of this attributes needs to be passed to the page:

`<div class="mimsib SmR8" jsname="EBHGs" data-challengeid=11 data-action=selectchallenge data-accountrecovery="false" data-challengetype="39">`

edwardrf:
Thanks @aaronthebaron, I did took a look at the page and tried to select the one-time code method, but it seems it always ends up sending me to the SMS challenge page. I'll have to circle back to see what could be the reason.
Code change here: https://github.com/edwardrf/saml2aws/commit/864417090e7dd537036519431c6bfb8836e1ad92

aaronthebaron:
@edwardrf Thanks for taking a swing at it. I fixed captcha's today and will take a look at your changes on Monday if you don't beat me to it.

aaronthebaron:
@edwardrf I think I see what's happening with your code. I think two elements are required, a `challengeId` and a `challengeType`. You have one but not the other. 

The types also appear to be in the HTML

```<div class="mimsib SmR8 RDPZE" jsname="EBHGs" data-challengeid=10 data-action=selectchallenge data-challengeunavailable="true" data-accountrecovery="false" data-challengetype="2">```

So it seems like you are on the right track?

mapkon:
What is the plan for this PR?

PR #1299, opened 2024-06-24
update error message to align with playwright-go update

charlottecampbell193:
New version of playwright-go 'run.go' file added a check on if the driver is up to date. Updated TestNoBrowserDriverFail in browser_test.go to look for the new error message.

```func Run(options ...*RunOptions) (*Playwright, error) {
	driver, err := NewDriver(transformRunOptions(options))
	if err != nil {
		return nil, fmt.Errorf("could not get driver instance: %w", err)
	}
	**up2date, err := driver.isUpToDateDriver()
	if err != nil || !up2date {
		return nil, fmt.Errorf("please install the driver (v%s) and browsers first: %w", playwrightCliVersion, err)
	}**
```


pkg/provider/browser/browser_test.go
```diff
@@ -102,7 +102,7 @@ func TestNoBrowserDriverFail(t *testing.T) {
 	client, _ := New(account)
 	_, err := client.Authenticate(loginDetails)
 	assert.Error(t, err)
-	assert.ErrorContains(t, err, "could not start driver")
```
mapkon, line 105: Thank you for this change @charlottecampbell193 - is there a scenario where the error "could not start driver" will be thrown? Or has that been completely removed?

mapkon (approved):
Minor clarification comment

pkg/provider/browser/browser_test.go
charlottecampbell193: @mapkon that error can still be thrown but not with the current test. A new test would need to be created that has the correct driver installed but has some other issue that is preventing it from starting. I'm not sure what scenario that would be though.

mapkon: You are right since the current error message is probably faulty, and `playwright` has fixed their API. The current test was exzercising the scenario where there is completely no driver, so it makes sense to throw the `please install the driver` message. The comment on the test gave it away:
`// Test that if download directory does not have browsers, it fails with expected error message`


I imagine that we should have a test that launches a directory with an existing driver that we know won't "start", and see if that will fire it up (probably an invalid browser binary or similar). I wouldn't spend more than a day on it if you cannot get it to start.

Lastly, remember to update the README on line 575 to reflect the new message that will get thrown if the user attempts use without installing the drivers (like on first time use).

charlottecampbell193: I was not able to recreate that error message. When I had an invalid browser binary the error I got was `'playwright: spawn UNKNOWN'`. Tried a bunch of stuff messing around with it and got many other unhelpful messages but not `could not start driver`.

README has been updated - let me know your thoughts on the update.

mapkon: Suffices for me. We can merge it now or wait for #1302 to be fixed (so we can run all the tests). Let me know

PR #1335, opened 2024-08-07
chore(docs): bump copyright year

tinaboyce:


mapkon (changes requested):
Lets try to make this [dynamic](https://stackoverflow.com/questions/54563526/is-there-markdown-syntax-that-will-allow-my-copyright-date-to-automatically-upda) so it is not missed in future?

PR #1342, opened 2024-09-06
Implement assume-time policy limiting

eldondevat:
When an STS token is acquired, it's possible to use supplemental policies to limit the permissions of the token. One example use might be assuming an administrative role, but limiting the permissions to a read-only or security-review scope to use to provided credentials in an auditing tool. This PR implements assume-time limiting of permissions with additional managed policies or a local policy document.

### DIFF
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -4,6 +4,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"strings"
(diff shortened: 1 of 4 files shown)


mapkon:
@eldondevat The linter is failing - can you take a look?

cmd/saml2aws/commands/login.go
```diff
@@ -365,6 +365,22 @@ func loginToStsUsingRole(account *cfg.IDPAccount, role *saml2aws.AWSRole, samlAs
...
+		}
+		params.Policy = aws.String(string(policy))
+	}
+
+	if account.PolicyARNs != "" {
+		var arns []*sts.PolicyDescriptorType
+		for _, arn := range strings.Split(account.PolicyARNs, ",") {
+			arns = append(arns, &sts.PolicyDescriptorType{Arn: aws.String(arn)})
+		}
+		params.PolicyArns = arns
```
mapkon, line 381: What happens when the policies/policyARNs are removed?

mapkon (commented):
Looks fine to me. I left a comment that needs clarification before merge.

cmd/saml2aws/commands/login.go
eldondevat: The policyfile would be a local file. If the local file doesn't exist, I would expect that to generate the new wrapped error. If the PolicyFile doesn't represent a valid policy, or if there is an attempt to use the arn of a policy that doesn't exist, I would expect AssumeRole to return an error, so those would end up with a wrapped error and "Error retrieving STS credentials using SAML." There are a few things we could do, like validating the format of ARN's or the general shape of the json policy document ahead of time, but I wasn't sure how much you would want to add there. I don't think we can validate the existence of the actual policies pointed to by the policyARN's in a meaningful way, because we don't really have AWS creds yet. I am using this while scoping things like AWS' canned ReadOnly and SecurityAudit policies, and those haven't gone away yet.
I wasn't sure what you meant by "removed" precisely, but hopefully it was covered in there.

PR #1386, opened 2025-01-24
expand `~` in saml_cache_file config field

rrotter:
currently setting a path like:
```
saml_cache_file  = ~/.aws/cache
```
results in the cache being created under a directory called `~` rather in $HOME.

This PR runs `homedir.Expand` on the configured path.


tinaboyce:
Thank you for your contribution! 🙇

Issue #1395, opened 2025-02-24
Keycloak authenticate via other brokers


nickbwuzhere:
We have a specific setup for our organization where we use multiple keycloak brokers to authenticate different regions of the organization. We are able to access these keycloak brokers when logging in through the browser by selecting the "login with ..." button below the login screen.

This poses a problem with our saml2aws implementation where anyone trying to authenticate against an AWS account using keycloak will have to use the browser mode in order to login. The browser mode supported by saml2aws is not the greatest, and is quite a clunky experience. It will not remember passwords, sessions, or allow extensions, causing quite a long logistical hurdle for logging in to an account in a different federation.

PR #1396, opened 2025-02-24
added kc broker flag

nickbwuzhere:
This pull request fixes the issues of saml2aws not having the ability to hop multiple request chains to follow keycloak brokers, as well as the issue of it not carrying cookies along with the requests in this chain, which would result in it failing to do any multi-provider hops. 

I mentioned the issue this is solving in #1395 

After looking at how the keycloak authentication chain operates, I've managed to add a flag called "kc-broker" for logging in. This flag you can give a valid kc-broker name, (specified in your keycloak configuration). Keycloak will then take this broker and follow through to the target broker, beginning the authentication chain. Once this process begins the only necessary component that needed adding to saml2aws was for saml2aws to carry cookies forward along the request chain. 

This avoids the usage of the browser mode, for requests that stay within the keycloak ecosystem, and allows usage of keychain to store credentials for accounts in other federations. 

It passes all tests, and works similar to other flags.  This could likely be expanded to other providers in the case the authentication has to go through several different sources to fully authenticate.

### DIFF
--- a/cmd/saml2aws/commands/login.go
+++ b/cmd/saml2aws/commands/login.go
@@ -25,7 +25,6 @@ import (
 
 // Login login to ADFS
 func Login(loginFlags *flags.LoginExecFlags) error {
-
 	logger := logrus.WithField("command", "login")
 
 	account, err := buildIdpAccount(loginFlags)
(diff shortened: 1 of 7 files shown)


tinaboyce:
Thank you for your contribution! 🙇

Issue #948, opened 2023-02-06
Keycloak login loop Status="200 OK" on bad password


attfuss:
# Keycloak login loop Status="200 OK" on bad password leads to problems with AD's security user locking

Hey together, i have an issue with saml2aws which maybe is related to scraping of the idp form login and it's interpretation. This issue is perceived when used with keycloak as an idp backed by an Active Directory (MS) as User Federation Provider.

used: saml2aws v2.36.2 with keycloak 20.0.1

## The use case: 
Login in with an false password, not using the linux keyring app. `saml2aws login`. 

## The outcome:
With This command this it the outcome:
```
λ saml2aws login --force --verbose
DEBU[0000] Running                                       command=login
DEBU[0000] Check if creds exist.                         command=login
DEBU[0000] Expand                                        name=/home/xx/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/home/xx/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/home/xx/.aws/credentials pkg=awsconfig
Using IdP Account default to access KeyCloak https://auth.xxx/auth/realms/xx/protocol/saml/clients/amazon-aws
To use saved password just hit enter.
? Username xx
? Password xx -> Int this case a false password to test the prerequisites

DEBU[0007] building provider                             command=login idpAccount="account {\n  URL: https://auth.xxx/auth/realms/rxx/protocol/saml/clients/amazon-aws\n  Username: xx\n  Provider: KeyCloak\n  MFA: Auto\n  SkipVerify: false\n  AmazonWebservicesURN: urn:amazon:webservices\n  SessionDuration: 36000\n  Profile: saml\n  RoleARN: \n  Region: eu-central-1\n}"
Authenticating as xx ...
DEBU[0008] HTTP Req                                      URL="https://auth.xxx/auth/realms/xx/login-actions/authenticate?session_code=XHS7ykoARgPFthna2KhUHyoae8lYGmtsB55U9hTW3hM&execution=0fe1a2f0-cf8a-4943-8de5-f6b1b84d6189&client_id=urn%3Aamazon%3Awebservices&tab_id=nBk9QquGBxI" http=client method=POST
DEBU[0008] HTTP Res                                      Status="200 OK" http=client
[...]

PR #1398, opened 2025-03-07
support invalid password for keycloak v2

nordicdyno:
support keycloak v2 in auth error validator

related to https://github.com/Versent/saml2aws/issues/948#issuecomment-2703840283


tinaboyce:
Thank you for your contribution 🙇

PR #1404, opened 2025-03-13
Improve error messages for invalid Keycloak credentials and OTP codes

nordicdyno:
prints:
- 'Invalid username or password.' instead of 'unable to locate saml response field' in case of password valiadation failed
- 'otp code is invalid'  instead of 'unable to locate saml response field' in case of otp code is invalid

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,9 +709,11 @@ http_retry_delay        = 1
 region                  = us-east-1
 ```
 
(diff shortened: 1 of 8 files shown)


nordicdyno:
what do you think about merging this to master? are any improvements needed, or there are reasons not to merge?

PR #1411, opened 2025-03-22
feat(CI): Create dependabot-tidy.yml

RossRichardson:
Ensure `go mod tidy` runs when Dependabot creates a PR

This ensures that dependabot Pull Requests are in a state where they can be built and tested succesfully as, often, a dependency update requires a go mod tidy to be run afterward.


RossRichardson:
The alternative is to separately update the go version in its own Pull Request and rerun the Dependabot PRs

RossRichardson:
Closed: In favour of a separate update of the go and go toolchain versions

PR #1418, opened 2025-03-24
fix: 🐛 improve password confirmation prompt (fixes #1390)

### LINKED ISSUE #1390: Improve "saml2aws configure" CLI prompting
TLDR: "Passwords did not match" is solved by entering the password again when prompted "Confirm" rather than pressing "Enter" to confirm.

When running the `saml2aws configure` command to login with Okta, the user is prompted `Confirm` after entering the password. This seems like a very unintuitive way to ask the user to input the same password again. It leads user to press "Enter" to confirm their inputs.

The error message "Passwords did not match" also doesn't make it explicit that the entered password and the confirmed one don't match (obviously because there isn't any confirmed password) and leaves the user perplexed, trying to understand which 2 passwords don't match.

Image

This can be very simply solved by changing the "Confirm" wording to something like "Confirm entered password" or "Enter password again", which is exactly what the CLI is trying to make the user do. An alternative would be to change the error message to "Confirmed password doesn't match".

I tried to open a pull request but I don't have the right permissions to do it. I hope this can be considered as a small but considerable improvement. Cheers!

srizzling:
Prior to this change:
- The saml2aws configure command included a password repeater for confirmation, which was unnecessary as this process is not a sign-up flow. If users entered an incorrect password, they could simply re-initiate the configuration process. The password repeater also caused ambiguity when passwords did not match, leading to a poor user experience.
- The console.go file contained a redundant error check for the err variable, which added unnecessary complexity to the code.

After this change:

- The password repeater has been removed entirely from the saml2aws configure command. This simplifies the flow and eliminates the ambiguity caused by the repeater. If users enter an incorrect password, they can now restart the configuration process without confusion.
- The redundant error check in console.go has been removed, simplifying the code and improving its readability.

fixes #1390

### DIFF
--- a/cmd/saml2aws/commands/configure.go
+++ b/cmd/saml2aws/commands/configure.go
@@ -19,11 +19,9 @@ const OneLoginOAuthPath = "/auth/oauth2/v2/token"
 
 // Configure configure account profiles
 func Configure(configFlags *flags.CommonFlags) error {
-
 	idpAccountName := configFlags.IdpAccount
 	idpAccountPassword := configFlags.Password
 
-	// pass in alternative location of saml2aws config file, if set.
 	cfgm, err := cfg.NewConfigManager(configFlags.ConfigFile)
 	if err != nil {
 		return errors.Wrap(err, "failed to load configuration")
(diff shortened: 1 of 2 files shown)


cmd/saml2aws/commands/configure.go
```diff
@@ -44,19 +44,23 @@ func Configure(configFlags *flags.CommonFlags) error {
...
+		if password == "" {
+			log.Println("No password supplied")
+			return nil
+		}
+
+		repeat := prompter.Password("Re-enter password")
+		if repeat != password {
+			log.Println("Confirmed password does not match")
+			os.Exit(1)
+		}
```
pwmcintyre, line 61: is this a good idea? 💭
```suggestion
		for {
			repeat := prompter.Password("Re-enter password")
			if repeat == password {
				break
			}
			log.Println("Confirmed password does not match")
		}
```

RossRichardson: Agree with repeat. For sign up: maybe? for logins: NO. let's delete?

srizzling: its one time during configure operation.. so I guess if a user gets a password wrong.. we are just assuming they will just do it again.

Seems okay to me.

ary-b (commented):
Thanks a lot for taking the time to look into this team, I really appreciate it 😄 I agree that the password repeater could be removed, but not a big deal if it stays with the enhanced prompts. Cheers everyone!

cmd/saml2aws/commands/configure.go
srizzling: I've squashed - and rebased (sorry @pwmcintyre) - and removed the password repeater.

I couldn't help myself interms of doing some refactoring... I tried to keep as minimal as possible.

RossRichardson (approved):
LGTM.
Acknowledging `Password` prompt now changed to `Enter Password`.

PR #1427, opened 2025-03-27
add support for first adapter

jsolares-nfl:
First adapter is basically a way to direct users to different idps in pingfederate based on user input, we're using email as the input for username, so first adapter prompts you for email, then based on domain you can get sent to okta, pingfederate , etc.

This makes pingid 2fa work when using first adapter.


jsolares-nfl:
https://github.com/Versent/saml2aws/discussions/1426

Issue #101, opened 2017-11-30
The --skip-prompt option should prompt for a password if not specified with --password


fgimian:
It is undesirable to use the `--password` option due to safety.  but `-skip-prompt` is at least useful to skip the Username prompt.  However, this is what happens:

```
fots > ~ $ saml2aws --skip-prompt -a fots login
Using IDP Account amp to access KeyCloak https://...
Authenticating as fots ...
Empty password
error validating login details
github.com/versent/saml2aws/cmd/saml2aws/commands.Login
        /Users/markw/go/src/github.com/versent/saml2aws/cmd/saml2aws/commands/login.go:65
main.main
        /Users/markw/go/src/github.com/versent/saml2aws/cmd/saml2aws/main.go:95
runtime.main
        /usr/local/Cellar/go/1.9.2/libexec/src/runtime/proc.go:195
runtime.goexit
        /usr/local/Cellar/go/1.9.2/libexec/src/runtime/asm_amd64.s:2337
```

This works as expected for the MFA token input with the `--skip-prompt` option:

```
fots > ~ $ saml2aws --skip-prompt --password=pass123 -a amp login
Using IDP Account amp to access KeyCloak https://...
Authenticating as fots ...

Security Token [000000]
:
```

Cheers
Fotis

chamila-c:
I think this would be a good opportunity to clarify the use cases for which `--skip-prompt` is intended.

One scenario where the current behaviour has been desirable is with automation scripts that need to run without human intervention: often in these cases, it's useful to have the command fail rather than hang on a prompt.

The other factor that may need consideration is the (admittedly, not all that well documented) ability  to use a `SAML2AWS_PASSWORD` environment variable to supply the password instead of the command line flag, making it possible to avoid prompting while also not having to use the `--password` option on the command line.

torbenkohlmeier:
I would also like to have an option to skip the username prompt. Sometimes I accidentally start typing my password when the username is prompted which is then visible in cleartext. This might happen while sharing my screen which would be really bad. Having an option to skip the username prompt would prevent this. This could even be the default behavior. Why are you prompted to enter a username if it is already given in the command line?

erpel:
For the browser based workflow, there is another complication: I prefer entering my password into the browser directly, so I use skip-prompt to make saml2aws take the username from config and prefill that in the browser but leave the password empty.

Modification to the skip-prompt behavior exactly as described here, would break that flow. 
[...]

PR #1428, opened 2025-03-28
Add `--try-no-prompt` flag

xenia-lang:
`--try-no-prompt` uses credentials from non-interactive sources, if possilbe.Otherwise it prompts the user to enter these credentials.

This optimizes the workflow by eliminating the need to press enter when valid credentials are present, while allowing to have them added to the keyring without the need to rerun the command (as would be necessary with `--no-prompt`). `--no-prompt` is still available for non-interactive usecases (see discussion in #101). This also creates an obvious place in the interface to add the functionality discussed in #101.

Naming of the new flag is open for discussion. Its placement (`loginFlags` vs `commonFlags`) is as well.

Unfortunately, I do not have an overview over all the use cases for saml2aws. If these changes hinder any use case, guidance is much appreciated. Same goes for general use of the Go language.


tinaboyce:
@xenia-lang Massive thanks for working on this. I have tested locally and can confirm it is working as expected, and I'm looking forward to using this.

xenia-lang:
Very happy to get this merged. Since I am also really looking forward to it, I'd like to ask: When is the next release planned?

To be clear, I'm not trying to be pushy. I'd just like to get an idea of how long we need to wait before we can use this.

PR #557, opened 2020-09-13
AWS partitions support for `saml2aws console`

duckfez:
Fixes #533. 

Adds support for `saml2aws console` to use `region=` in config file 
as a hint for which partition to open the console in.

PR #1429, opened 2025-03-31
Support console subcommand with alternate federation urls

### LINKED ISSUE #533: govCloud requires alternate federation URL
Hello,

I am attempting to use `saml2aws console` to login directly to the AWS console after doing the authentication dance, however it is taking me to AWS Commercial for the federation endpoint, rather than to the govCloud endpoint.

`https://signin.amazonaws-us-gov.com/federation`

I have set my region to `us-gov-west-1` and that works fine for logging in (the default does not work because once again it tries to exchange the SAML assertion at the wrong STS endpoint), but it seems that the federation URL is hardcoded and is not switched depending on the environment.

### LINKED ISSUE #905: [BUG] Browser login does not support GovCloud

jolexa:
Fixes #533
Fixes #905

Hello, following up from PR #557 -- I have successfully tested this change with a GovCloud account. I _think_ the only hold up on #557 was access to a govcloud aws account as it seemed that PR was ready for inclusion (pending rebase).

I'm not sure if I love this approach using region versus setting an explicit federation url but that would be a much more invasive change that I don't feel confident in. Thanks for considering!

### DIFF
--- a/cmd/saml2aws/commands/console.go
+++ b/cmd/saml2aws/commands/console.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


jolexa:
@tinaboyce (Apologies for pinging you directly) any feedback? Asking because you last merged a PR in this repo. Any response is greatly appreciated

jolexa:
I've abandoned supporting this PR so I'll close it off.

PR #1431, opened 2025-04-05
feat: Route --version output to stdout

neo-anderson:
Currently, `saml2aws --version` prints the application version to `stderr`. This change modifies the behavior to print the version string to `stdout` instead while other logs and errors remain on `stderr`.

**Reasoning:**

Printing the version to `stdout` aligns better with common scripting practices and standard CLI tool behavior for version reporting. This makes it easier to capture the version string programmatically without needing `stderr` redirection (`2>&1`).

For example, users with bootstrapping or environment setup scripts that check dependency versions currently need to add special handling for `saml2aws` like
```
version = result.stdout.strip() if tool != "saml2aws" else result.stderr.strip()
```
or like 
```
VERSION=$(saml2aws --version 2>&1)
```
Routing `--version` output to `stdout` removes this inconsistency.

https://github.com/Versent/saml2aws/issues/1432


RossRichardson (changes requested):
Instead of reimplementing the `Version()` function it looks like we can set the usage and version output to write with `os.stdout` using `.UsageWriter()`.

So we keep the `.Version(` and we set the `.UsageWriter(`. I've checked, but we should verify, this should not interfere with error messages caused by incorrect commands from going to stderr because that is handled by `fatalUsage()`.

https://github.com/alecthomas/kingpin/blob/64d23c54e3e2a385ce12757939bed9c632d1693c/app.go#L146

neo-anderson:
> Instead of reimplementing the `Version()` function it looks like we can set the usage and version output to write with `os.stdout` using `.UsageWriter()`.
> 
> So we keep the `.Version(` and we set the `.UsageWriter(`. I've checked, but we should verify, this should not interfere with error messages caused by incorrect commands from going to stderr because that is handled by `fatalUsage()`.
> 
> https://github.com/alecthomas/kingpin/blob/64d23c54e3e2a385ce12757939bed9c632d1693c/app.go#L146

Thanks for the review. That's way cleaner.

neo-anderson:
I verified that errors are sent to stderr by running this: `$GOPATH/bin/saml2aws invalid 1>out.txt 2>err.txt`

Issue #1072, opened 2023-06-06
Authentication on Azure AD with MFA doesn't work, keeps asking MFA code


igordust:
As per the subject, the authentication with Azure AD with MFA enabled doesn't work, it's asking the MFA code indefinitely. I attach a debug session.

```
$ saml2aws --version
2.36.6

saml2aws login --verbose
DEBU[0000] Running                                       command=login
DEBU[0000] Check if creds exist.                         command=login
DEBU[0000] Expand                                        name=/Users/zzzzz/.aws/credentials pkg=awsconfig
DEBU[0000] resolveSymlink                                name=/Users/zzzzz/.aws/credentials pkg=awsconfig
DEBU[0000] ensureConfigExists                            filename=/Users/zzzzz/.aws/credentials pkg=awsconfig
Using IdP Account default to access AzureAD https://account.activedirectory.windowsazure.com
DEBU[0000] Get credentials                               helper=osxkeychain serverURL="https://account.activedirectory.windowsazure.com"
DEBU[0000] Get credentials                               helper=osxkeychain user=zzzzzz
To use saved password just hit enter.
? Username zzzzzz
? Password

DEBU[0001] building provider                             command=login idpAccount="OMIT"
Authenticating as zzzzzz ...
DEBU[0002] processing ConvergedSignIn                    provider=AzureAD
DEBU[0002] HTTP Req                                      URL="https://login.microsoftonline.com/common/GetCredentialType?mkt=en-US" http=client method=POST
DEBU[0002] HTTP Res                                      Status="200 OK" http=client
DEBU[0002] HTTP Req                                      URL="https://login.microsoftonline.com/common/login" http=client method=POST
DEBU[0002] HTTP Res                                      Status="200 OK" http=client
DEBU[0002] processing a 'hiddenform'                     provider=AzureAD
DEBU[0002] HTTP Req                                      URL="https://device.login.microsoftonline.com:443/" http=client method=POST
DEBU[0003] HTTP Res                                      Status="200 OK" http=client
DEBU[0003] processing a 'hiddenform'                     provider=AzureAD
DEBU[0003] HTTP Req                                      URL="https://login.microsoftonline.com:443/common/DeviceAuthTls/reprocess" http=client method=POST
DEBU[0003] HTTP Res                                      Status="200 OK" http=client
DEBU[0003] processing ConvergedTFA                       provider=AzureAD
[...]

PR #1468, opened 2025-09-22
Adding AAD CMSI Interrupt page processing capability - Fix Issue 1072

beardtm:
Adding AAD CMSI Interrupt page processing capability and adding missing parameters for other pages in AAD authentication. Fixes issue 1072.


icirellik:
Thank you for identifying the missing ProcessAuth fields. I opened #1538 as a narrower follow-up containing only those continuation fields, compatibility guards, and regression coverage.

We reproduced #1072 in a real Entra MFA flow: the original client repeated MFA after Microsoft accepted the code, while the focused patch completed sign-in in the same environment. I referenced and credited this PR, and kept the CMSI/KMSI changes separate to make review easier.

PR #1471, opened 2025-10-09
feat(pingone): handle device profiling form

estahn:
The PingOne provider stopped working today most likely due to device profiling methods being enabled. The PR adds an additional check and form submission for this case.

See "[PingOne Integrations > PingOne Protect Integration Kit > Device profiling methods](https://docs.pingidentity.com/integrations/pingone/pingone_protect_integration_kit/pf_p1_protect_ik_device_profiling_methods.html)" for more details.


estahn:
@mapkon Mark, would you mind merging this?

davidobrien1985:
@estahn ,  @tinaboyce might be better to tag here. I don't know if anybody is still actively maintaining this since @wolfeidau left Versent.

PR #1484, opened 2025-10-28
fix(security): use proper uuid for device token

reegnz:
Okta device tokens are supposed to be unguessable. They are meant to help in not having to MFA constantly for devices, essentially a short-cirtuit to avoid needing MFA. So in sensitivity they are between your password and your MFA, therefore they are NOT meant to be precomputed from the username (as saml2aws does it today).

If the device-token is easily guessable, then it defeats MFA, and the current users of saml2aws live in a false sense of security when they allow remember-my-device (the default with saml2aws).

With saml2aws using a predictable device token, it poses a security risk for any user using saml2aws, as it defeats using MFA.

This fix aims at generating a UUID specific to the device saml2aws runs on and storing it as a secret (like storing the users password). This makes using 'remember device' safe.


reegnz:
@tinaboyce I didn't find your contact info anywhere besides linkedin but I could not send you a message on LinkedIn because it's a paid feature 🙃 .

I tried finding a proper channel to signal security issues, but couldn't find one. Tried reaching out to the top contributor, to no avail, didn't get a response in 2 weeks. I see he's not even a versent employee anymore.

Anyway, this is relatively mild security hole IMHO, not immediately exploitable, but it significantly reduces the security posture of a user that uses MFA, they think they're safe, but the current state of device token handling in saml2aws completely defeats MFA, so you're only relying on password strength.

PR #1491, opened 2025-11-20
Support custom user agent

### LINKED ISSUE #1255: Custom UserAgent to Identify saml2aws as a desktop app
It would be very helpful if the saml2aws identifies itself as a Desktop app through browser user agent string to EntraID so separate policies can be defined for saml2aws app as a desktop app. This will help define non FIDO2 requirement policy for AWS admin login until a FIDO2 support is enabled in saml2aws.

Cyb3r-Jak3:
Adds two new options to either append or override the user-agent that is sent. Should close #1255

I also moved the version to its own pkg so that it can be referenced in multiple places.


dosmanak:
Why was not it merged?

PR #1509, opened 2026-01-23
Add browser cookies in keyring

soham-sl:
This update enables the Browser provider to save cookies in the system keyring so subsequent logins can reuse them. It follows the logic used for Okta sessions: if `disable_cookies` is set in the config or the `--disable-keychain` argument is used, it will not read or write any cookies.

### DIFF
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -147,8 +147,8 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	}
 
 	defer func() {
-		logger.Info("saving storage state")
 		if cl.DisableCookies {
+			logger.Info("saving storage state")
 			cookies, err := context.Cookies(loginDetails.URL)
 
 			if err != nil {
--- a/pkg/provider/browser/browser.go
+++ b/pkg/provider/browser/browser.go
@@ -1,17 +1,19 @@
 package browser
 
 import (
+	"path"
 	"errors"
 	"fmt"
 	"net/url"
-	"os"
 	"regexp"
 	"strings"
+	"encoding/json"
 
 	"github.com/playwright-community/playwright-go"
 	"github.com/sirupsen/logrus"
 	"github.com/versent/saml2aws/v2/pkg/cfg"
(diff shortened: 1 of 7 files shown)


pkg/provider/browser/browser.go
```diff
@@ -113,20 +115,25 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
...
 		return "", err
 	}
-	if _, err := os.Stat(storageStatePath); err == nil {
-		contextOptions.StorageStatePath = playwright.String(storageStatePath)
+
+
+    if loginDetails.CookiesJson == "" {
+        logger.Info("could not retrieve cookies")
+    } else {
+		logger.Info("Cookie payload bytes: ", len(loginDetails.CookiesJson))
```
bhavya-sl: Either this print statement should say the "length of the cookies in string", or change it to print `len([]byte(loginDetails.CookiesJson))`

pkg/provider/browser/browser.go
```diff
@@ -136,7 +143,16 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 
 	defer func() {
 		logger.Info("saving storage state")
-		_, err := context.StorageState(storageStatePath)
+		cookies, err := context.Cookies()
```
bhavya-sl: Can we specify any URL here? So that we don't fetch all the cookies in `Authenticate`
Below taken from `playwright-go`
```
	// If no URLs are specified, this method returns all cookies. If URLs are specified, only cookies that affect those
	// URLs are returned.
	Cookies(urls ...string) ([]Cookie, error)
```

pkg/provider/browser/browser.go
```diff
@@ -135,10 +147,24 @@ func (cl *Client) Authenticate(loginDetails *creds.LoginDetails) (string, error)
 	}
 
 	defer func() {
-		logger.Info("saving storage state")
-		_, err := context.StorageState(storageStatePath)
-		if err != nil {
-			logger.Info("Error saving storage state", err)
+		if cl.DisableCookies {
```
bhavya-sl: Should this be `!cl.DisableCookies`?

